PR #5, opened 2024-05-09
CLP-10747 Character styles don't trump paragraph styles

jurisdatum:
In MS Word, the formatting of a character style does not trump the character formatting of a paragraph style.

For example, if there is a paragraph style named `ParaHeading1` with bold text, and a character style named `fontstyle01` with normal (not bold) text, and if this character style is associated with a run of text within a paragraph associated with this paragraph style, then the text within that run will be bold. (This example is taken from [2024] UKSC 14, included as test/judgments/test82.docx.)

However, in Akoma Ntoso (and HTML), the formatting of a class associated with an inline element _does_ trump the character formatting of the class associated with an enclosing block element. Therefore, if we represent our run in Akoma Ntoso like so:
```xml
<p class="ParaHeading1"><span class="fontstyle01">The Issues</span></p>
``` 
the implied formatting of the enclosed text would _not_ be bold.

To compensate, the parser will now produce the following XML:
```xml
<p class="ParaHeading1"><span class="fontstyle01" style="font-weight:bold">The Issues</span></p>
```


jurisdatum (commented):
reviewed

jurisdatum (commented):
Leave a comment

TomJKing (approved):
Approved.

PR #48, opened 2025-02-12
Add .gitignore file

anthonyhashemi:
Add .gitignore file to ignore files we do not want to track in the repo. Adding the files that I have got after running the parser in the directory. Will expand this as we develop.


jurisdatum:
thanks

PR #54, opened 2025-03-08
enrich from beginning

jurisdatum:
This helper function enriches text patterns from the _beginning_ of a line, preserving formatting variations that might follow. Useful for definitions in legislation.


mcpheed1:
Thanks Jim! I finally got around to testing these changes. 

I noticed that The Adoption and Children Bill (`test4.docx`) began failing after pulling the changes from this branch into the `LNI-240` branch. The problem occurred at this italic semicolon at the end of the following definition in Section 158. Presumably this italicisation is a drafting error, but oh well. 
Definition
The output shows the inlines situated in reverse order -
```
<hcontainer name="definition" class="definition">
    <content>
        <p>;<def ukl:startQuote="“" ukl:endQuote="”">court</def> has the meaning given by section 115(1)</p>
    </content>
</hcontainer>
```
Simply removing the call to  `.Reverse()` in the following line was enough to fix the issue. 
```
IEnumerator<IInline> enumerator = raw.Reverse().GetEnumerator();
```

I then started thinking about the possibility of inline styles which are either wholly or partially inside the defined term. For example:
image
These inline styles were being removed, as the `Contents` of the resulting `Def` element were being constructed using plain text. We have a ticket in our next sprint for respecting inline styles in the Bill parser, and I found myself with some downtime today, so I took a stab at handling the above scenarios. 
The solution involved constructing the `Def` element from a list of `IInline` objects rather than plain text, and ensuring that any `Inline` objects which cross the boundary of the defined term are split , such that only the portion inside the defined term is included in the `Contents` of the `Def`, while preserving any inline styles. I've also added a test case derived from the image shown above.
I'd be keen to hear your thoughts on this change. It's quite a substantial departure from the original logic, sorry. Though I'd be happy to revert my changes if you're not a fan.

PR #58, opened 2025-03-20
LNI-251: Handle remaining issues with Quoted Structure handling

mcpheed1:
- Added missing Quoted Structure attributes: `ukl:indent="indent0"`, `ukl:docName="nia"`, and `ukl:context="body"`
- Allowed section/subsection numbers with multiple strings of digits separated by letters (i.e. '63Z1' from the Justice Bill)
- Allowed Grouping Provisions as Leaves (these can occur in Quoted Structures)
- Relaxed italicisation requirement for Cross Headings. Instead of the whole line needing to be italic, only part of the line need be so. This is so that non-italic quote marks and/or appendTexts surrounding quoted Cross Headings don't prevent the parser from recognising them as Cross Headings.
- Removed all-caps requirement for Schedule Numbers for the same reason. Also allowed Schedule Numbers to contain letters, like Grouping Provisions. 
- Updated the ExtractQuotedAndAppendTexts algorithm such that it works when the end quote / appendText spans multiple Inlines.


mcpheed1:
Closing this pull request without merging. These changes will instead by included in the upcoming `sprint-23` PR, per our discussion earlier this week about transitioning to staggered fortnightly PRs.

PR #61, opened 2025-03-31
Sprint 23

mcpheed1:
Sprint 23 comprised 3 tickets:

**LNI-251**
_Address remaining issues with Quoted Structure handling_
 - Added missing Quoted Structure attributes: ukl:indent="indent0", ukl:docName="nia", and ukl:context="body"
 - Allowed section/subsection numbers with multiple strings of digits separated by letters (i.e. '63Z1' from the Justice Bill)
 - Allowed Grouping Provisions as Leaves (these can occur in Quoted Structures)
 - Relaxed italicisation requirement for Cross Headings. Instead of the whole line needing to be italic, only part of the line need be so. This is so that non-italic quote marks and/or appendTexts surrounding quoted Cross Headings don't prevent the parser from recognising them as Cross Headings.
 - Removed all-caps requirement for Schedule Numbers for the same reason. Also allowed Schedule Numbers to contain letters, like Grouping Provisions.
 - Updated the ExtractQuotedAndAppendTexts algorithm such that it works when the end quote / appendText spans multiple Inlines.

**LNI-247**
_Assign class="unknownImport" to unidentified paragraphs_
 - This class is added to unidentified paragraphs so that the Oxygen editor in the Lawmaker application can apply special formatting to them, making them easier for drafters to spot.  

**LNI-246**
_Implement manual quoted structure 'coding'_
 - Added support for document types other than `NIA`...
    - Added the `Frames` class, which maintains a stack of `Frame` objects.
    - An individual `Frame` comprises a `DocName` (i.e. `NIA`, `UKPGA`, `NISI`) and a `Context` (i.e. `BODY`, `SCH`).
    - Whenever we enter a Quoted Structure, a new `Frame` is added to the stack, potentially with a different `DocName` and/or `Context`. 
    - When we exit the Quoted Structure, this frame is popped off the stack. Hence the top of the stack always reflects the active  `DocName` and `Context` at the current location in the document.
    - Something similar happens whenever we enter a schedule - a new `Frame` is added with the `Context` of  `SCH`, however the `DocName` remains unchanged.
    - Inside a given `Parse()` method (i.e. `ParseCrossHeading()`), the `Frame` at the top of the stack is consulted, and slightly different parsing logic might execute depending on its `DocName` and `Context`.
- Allowed users to manually dictate the `DocName` and `Context` of a given quoted structure.
   -  This is done by entering a 'code' in braces before the start quote. 
   - These 'codes' are ultimately stripped from the plain-text and end up in the `ukl:docName` and `ukl:context` attributes of the resulting quoted structure. 
   - The format is `{DocName-Context}`, where the `Context` is optional (defaults to `BODY` for primary legislation)
image
   - For Statutory Instruments in particular, there is a further need to distinguish between Rules, Regulations, and Orders in the Body, so the full set of possible `Context` values is actually {`BODY`, `RULES`, `REGS`, `ORDER`, `SCH`}. When not specified, the default `Context` is `REGS` for secondary legislation.
   - Some example codes:
       - `{UKPGA}`
       - `{UKPGA-BODY}`
       - `{UKPGA-SCH}`
       - `{ASP}`
       - `{NISI}`
       - `{NISR-REGS}`
       - `{SSI-RULES}`
       - `{UKSI-ORDER}`
   - If a quoted structure has a code, but the `DocName` or `Context` (or both) are not recognised as valid options, default values are assumed, and the Quoted Structure will be assigned `class="unknownImport"`


jurisdatum:
well done

PR #69, opened 2025-04-20
Move 'unknownImport' class to lawmaker package

jurisdatum:
@mcpheed1 @chatawayf I couldn't figure out how to work around the UnkownLine entirely. I thought they were only ever children of an UnknownLevel, but then I saw that they can be top-level children of a Mod, obviously.

So, as a compromise, I've proposed moving the bit about adding the "unknownImport" class attribute from the root Builder to the Lawmaker builder. My thinking is that the meaning of that class seems specific to Lawmaker, and unlikely to be used anywhere else.

### DIFF
--- a/src/akn/Builder.cs
+++ b/src/akn/Builder.cs
@@ -375,19 +375,10 @@ abstract class Builder {
             AddInline(block, inline);
         ContainingParagraphStyle = null;
     }
-    protected void p(XmlElement parent, ILine line) {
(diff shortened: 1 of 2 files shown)


mcpheed1:
Thanks @jurisdatum, this change makes sense. The Lawmaker unit tests pass, and I tested other Bills manually with no issue. I'll go ahead and merge.

mcpheed1 (approved):

PR #72, opened 2025-07-11
LNI-306: set default ns for direct child of td

chatawayf:
For better or worse, Lawmaker sets the default namespace for direct child elements in table cells *back* to AKN to prevent the need to specify the akn namespace for any structured content its children.

This commit sets this xmlns attribute on all direct children of a cells content.


mcpheed1:
A quick and easy change to account for Lawmaker's less-than-perfect handling of namespaces in tables. Subject to change when we eventually get around to handling structured content in table cells.  LGTM!

PR #94, opened 2025-07-28
Improve handling of tables of contents within SDT blocks

jurisdatum:
There was already some logic for these cases, but it was invoked only during the main parsing phase. This fix extends it to the pre-parsing phase too.

Update[from @anthonyhashemi ]: undone the hacky fix we implemented temporarily to get the doc we saw this fail in parsing, and added test to cover this case.


anthonyhashemi:
Spoken with Jim on slack, and since this fix is not urgent now we have the temporary one working, we will take some more time to make sure we do not have duplicate/redundant code now. And will add tests around this. He has a lot else on so I will finish these last bits

jacksonj04 (approved):

PR #117, opened 2025-08-01
Generalise JUDGMENTS_FILE_PATH and HMCTS_FILE_PATH

anthonyhashemi:
Generalise JUDGMENTS_FILE_PATH and HMCTS_FILE_PATH since different tribunal backlog uploads may have different path prefixes in the csv metadata files.


backlog/src/Program.cs
```diff
@@ -29,6 +29,9 @@ public static int Main(string[] args)
 
                 DotNetEnv.Env.Load();  // required for bucket name
 
+                string judgmentsFilePath = Environment.GetEnvironmentVariable("JUDGMENTS_FILE_PATH");
```
dragon-dxw, line 32: How many of these are we going to have? Do we need a more sustainable model?

anthonyhashemi: as in how many env vars or how many potentially different judgmentsFilePath values from the different tribunals?

Env vars: there may be some other things that are hardcoded and we want to generalise but hopefully not many more

different judgmentsFilePath from tribunals: potentially lots but this means we are agnostic to that

anthonyhashemi: will continue discussion on slack

PR #129, opened 2025-08-06
Simplify csv metadata validation

anthonyhashemi:
Simplify csv metadata validation.
Realised that the csvreader has its own inbuilt validation and so use [Optional] attribute in lines class

### DIFF
--- a/test/backlog/TestHelperFindLines.cs
+++ b/test/backlog/TestHelperFindLines.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using NUnit.Framework;
 using Backlog.Src.Batch.One;
 using CsvHelper;
+using CsvHelper.Configuration;
 
 namespace Backlog.Test
 {
@@ -14,7 +15,6 @@ namespace Backlog.Test
         private Helper helper;
         private string testDataDirectory;
         private string validCsvPath;
-        private string invalidCsvPath;
         
         [SetUp]
         public void Setup()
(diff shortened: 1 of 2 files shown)


test/backlog/TestHelperFindLines.cs
```diff
@@ -129,30 +125,6 @@ public void FindLines_WithZeroId_ReturnsEmptyList()
             Assert.That(result.Count, Is.EqualTo(0), "Should return empty list for ID 0");
         }
 
-        [Test]
-        public void FindLines_WithValidationError_ThrowsInvalidOperationException()
```
anthonyhashemi, line 133: redudnant, we have tests around missing column headers below

test/backlog/TestHelperFindLines.cs
```diff
@@ -59,15 +55,6 @@ private void CreateValidCsvFile(string path)
             File.WriteAllText(path, csvContent);
         }
 
-        private void CreateInvalidCsvFile(string path)
```
anthonyhashemi, line 62: and this remove for same reason

test/backlog/TestHelperFindLines.cs
```diff
@@ -27,10 +27,6 @@ public void Setup()
             validCsvPath = Path.Combine(testDataDirectory, "valid-metadata.csv");
             CreateValidCsvFile(validCsvPath);
             
-            // Create invalid CSV file missing required columns
-            invalidCsvPath = Path.Combine(testDataDirectory, "invalid-metadata.csv");
```
dragon-dxw, line 31: Is this a file that can be deleted?

PR #130, opened 2025-08-06
Require CaseNo column, rather than separate constituent number columns

anthonyhashemi:
Require CaseNo column, rather than separate constituent number columns


dragon-dxw (approved):
I assumed there was a good reason for doing it the other way, this way is much better. :D

anthonyhashemi:
@dragon-dxw yeah I think it was just because the first examples had it like that and we initially were going to have very minimal editor input but now we are preprocessing the csvs anyway, preprocessing them to be in this format is ok

PR #131, opened 2025-08-06
Read in NCNs when available in tribunal metadata csv

anthonyhashemi:
https://national-archives.atlassian.net/browse/FCL-1041

Read in NCNs when available in tribunal metadata csv

### DIFF
--- a/backlog/README.md
+++ b/backlog/README.md
@@ -93,6 +93,7 @@ The following columns are optional:
 - `main_subcategory` - Primary subcategory name (child of main_category)
 - `sec_category` - Secondary category name (optional)
 - `sec_subcategory` - Secondary subcategory name (child of sec_category, only used if sec_category is provided)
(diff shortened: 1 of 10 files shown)


backlog/README.md
```diff
@@ -93,6 +93,7 @@ The following columns are optional:
 - `main_subcategory` - Primary subcategory name (child of main_category)
 - `sec_category` - Secondary category name (optional)
 - `sec_subcategory` - Secondary subcategory name (child of sec_category, only used if sec_category is provided)
+- `ncn` - Neutral Citation Number (NCN) for the judgment, when available. If provided, this appears as `uk:summaryOfCite` in the generated AkomaNtoso XML
```
dragon-dxw, line 96: this should only be true of press summaries; it should be uk:cite in normal judgments/decisions

anthonyhashemi: my bad, I think I looked at a press summary when looking for an example 🤦

PR #133, opened 2025-08-06
Handle appellants or claimants as FirstParty

anthonyhashemi:
Handle appellants or claimants as FirstParty


backlog/src/batch/one/Metadata.cs
```diff
@@ -39,7 +40,7 @@ public override bool IsValid(object value)
 
         public override string FormatErrorMessage(string name)
         {
-            return "Subcategory columns can only exist if their main category is defined";
+            return "Subcategory columns can only exist if their main category is defined, and exactly one of claimants or appellants must be provided";
```
dragon-dxw, line 43: These feel like very different errors, but I don't think it's super important we separate them.

PR #137, opened 2025-08-07
LNI-320

chatawayf:
So far, the support for Lawmaker documents has been restricted to NI Public Bills to simplify implementation. This PR adds support to the parser to handle take in and handle more types of legislation that Lawmaker will supply.
This does include some changes to non-Lawmaker code (see `Program.cs`) to supply the necessary parameters locally though I'm unsure if this is the best approach, it seems a bit ugly to me.

The changes supply the correct `DocName` is available in the `Frame` during parsing, allowing for conditional logic during the parsing and building.

### DIFF
--- a/src/lawmaker/models/Document.cs
+++ b/src/lawmaker/models/Document.cs
@@ -6,10 +6,10 @@ using UK.Gov.Legislation.Judgments;
 namespace UK.Gov.Legislation.Lawmaker
 {
 
-    public abstract class Bill
+    public class Document
     {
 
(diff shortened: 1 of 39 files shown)


src/lawmaker/models/Document.cs
```diff
@@ -6,10 +6,10 @@
 namespace UK.Gov.Legislation.Lawmaker
 {
 
-    public abstract class Bill
+    public class Document
```
chatawayf, line 9: Not at all sold on this name, open to suggestions

PR #141, opened 2025-08-07
Read in court from tribunal metadata csv

anthonyhashemi:
Read in court from tribunal metadata csv

Reasoning: otherwise too difficult to manage for different tribunal batches. Could alternatively had done an input artgument or env var, but since some batches so far have 2 courts in 1 spreadsheet, this keeps things simple even though practically most will just ahve one court value in all rows.

### DIFF
--- a/test/backlog/test-data/EmptyCSVTest/court_metadata.csv
+++ b/test/backlog/test-data/EmptyCSVTest/court_metadata.csv
@@ -1,1 +1,1 @@
-id,decision_datetime,CaseNo,claimants,respondent,main_category,main_subcategory,Name,FilePath,Extension
+id,court,decision_datetime,CaseNo,claimants,respondent,main_category,main_subcategory,Name,FilePath,Extension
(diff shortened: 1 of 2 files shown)


backlog/README.md
```diff
@@ -122,6 +119,26 @@ bulk_num,trib_id
...
+
+Examples of valid court codes include:
+
+- `EWHC-QBD-Admin` - Administrative Court (Queen's Bench Division)
+- `EWHC-Chancery` - Chancery Division  
+- `EWHC-Family` - Family Division
+- `UKFTT-GRC` - First-tier Tribunal (General Regulatory Chamber)
+- `UKIST` - Immigration Services Tribunal (pre-2010)
+
+### Court Code Validation
```
dragon-dxw, line 136: This section feels more important than the examples, maybe move it up (the "this is where they're defined" is particularly important)

dragon-dxw:
What's the intention for courts once they're uploaded? That this process is never run again, so the code can get out of date/versions of CSVs can change? Do we need some guidance around this?

anthonyhashemi:
> What's the intention for courts once they're uploaded? That this process is never run again, so the code can get out of date/versions of CSVs can change? Do we need some guidance around this?

Not too sure I understand, but will followup in slack

PR #166, opened 2025-08-20
Fix judgments tests in lawmaker branch

jurisdatum:
## Update tests and adjust alphabetic numbered paragraph detection patterns

This PR makes two adjustments to numbered paragraph detection after the lawmaker branch integration:

### Test output adjustments
Updated 8 judgment test files to reflect improvements to the <num> markup.

### Refined regex patterns for alphabetic numbering
The lawmaker branch introduced support for multi-letter paragraph numbering (e.g., `(AA)`, `(BB)`) by changing `[A-Z]` to `[A-Z]+` in the HardNumbers class. However, this was too permissive for judgments and incorrectly matched role labels like `(CLAIMANT)` and `(DEFENDANT)` in test 8.

**Changes made:**
- Limited alphabetic patterns to `{1,3}` characters instead of unlimited `+`
- Added specific pattern `[ivx]+[a-z]` to handle legislative amendment numbering like `(iiia)`. such as in lawmaker/test4.xml

### DIFF
--- a/src/parsers/optimized/HardNumbers.cs
+++ b/src/parsers/optimized/HardNumbers.cs
@@ -65,8 +65,8 @@ class HardNumbers {
 
     public static readonly string[] NumberFormats = new string[] {
         @"\d+\.",              @"\(?\d+\)",
-        @"[A-Z]\.",            @"\(?[A-Z]+\)",
-        @"[a-z]\.",            @"\(?[a-z]+\)",
+        @"[A-Z]\.",            @"\(?[A-Z]{1,3}\)",
+        @"[a-z]\.",            @"\(?[a-z]{1,3}\)",
         @"[ivx]+\.",           @"\(?[ivx]+\)",
         // compound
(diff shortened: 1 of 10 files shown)


jurisdatum:
@mcpheed1 Thanks so much for your thorough explanation.

My instinct is to try (as least initially) to be limited. So, I'd be in favor of trying a specific rule for the "z" prefixes, in part because I think it can serve as helpful documentation of your system, especially with an explanatory comment. And if you need more than one letter following a lower Roman numeral, perhaps we could try something like `[ivx]+[a-z]{1,3}`.

But this just be my lawyerly impulse to proceed incrementally and catch the failures when they arise. If it seems like too much trouble and you'd prefer to create a separate set of numbering formats specifically for Lawmaker, please go ahead.

cc @chatawayf

anthonyhashemi:
> @anthonyhashemi Do you want to have a look at this?

@jurisdatum I had already skimmed this and generally looks fine from fcl perspective, the changes to the test look like good enhancements 👌

jurisdatum:
I'm planning to merge this tomorrow

PR #167, opened 2025-08-25
LNI-317: Adjust logic for handling tabbed authorialNotes

erciyass:


mcpheed1:
**Reviewed**

- The bounds of signatures & signature blocks are consistently identified correctly, with one slight caveat:
  - When there are multiple adjacent signatures, the earlier signature often consumes the first left-aligned portion of the later signature.
    image
  - To be addressed in an upcoming ticket. 
  - Use of styles will help. I also wonder if we can rely on the presence of one or more newline characters between signatures, so long as there are never newline characters within signatures.
- The case where schedule reference notes are on the same line as the schedule num (in Statutory Instruments ) has been handled.
- The case where schedule reference notes are centre aligned (in Scottish Bills/Acts) has been handled.
- Tests pass.

PR #168, opened 2025-08-25
Sprint 33

mcpheed1:
**LNI-320**
**_Handle Statutory Instrument document types_**

Historically, only NI Bills could be imported by the parser. While we _had_ previously handled other document types specifically inside quoted structures, one could not import an entire Statutory Instrument, for example. This ticket adds support to the parser to take in and handle more types of legislation that Lawmaker will supply (see the `DocName` enum in `LegislationClassifier.cs` for the full list of supported document types). Note that this ticket _does_ include some changes to non-Lawmaker code (see `Program.cs`) to supply the necessary parameters locally. Francis was not 100% confident that this was the best approach, and is of course open to feedback.

We previously implemented a stack of `Frames` which would record the the `DocName` and `Context` at the current point in the document, given the presence of quoted structures. Originally, the default `DocName` to be used when _not_ inside a quoted structure was hard-coded as `NIPUBB`. Now, this value is populated with the `DocName` provided to the parser in the request.

**LNI-317**
**_Handle SI provisions_**

- Added the capability to parse signatures and signature blocks.
- Prevented extra white space between the nums of `Prov1` elements and their first `Prov2` child from causing the provisions to be misidentified.
  image
- Handled schedule authorialNotes which are not on their own separate line, but rather are right-aligned on the same line as the schedule num i.e.
  image

**LNI-318**
**_Parse the preface/preamble of SIs_**

The logic in `Header.cs` for populating the preface and preamble of Word documents was originally implemented with only Primary Legislation in mind. The preface and preamble of Secondary Legislation comprise altogether different content, which resulted in the parser failing to correctly determine their bounds.

This ticket is the first step toward fully parsing the preface and preamble—
 - The bounds of the preface, preamble and body of SI documents can now consistently be determined by the parser, informed by formatting cues.
 - The contents of the preamble are parsed and should be present in the output.
 - The contents of the preface are currently ignored. For now, the preface is absent from the output. Proper handling of the preface (and associated metadata) will be performed in an upcoming ticket.
 - Also expanded the testing infrastructure to handle arbitrary doc types. Test files are placed in sub-directories named according to their doc type. The names of these sub-directories inform the parser of the doc type.

**LNI-279**
**_Handle footnotes_**

This ticket serves to fix the slightly incorrect formatting of parsed footnotes in imported Lawmaker documents:
 - The class attribute was being set, but was then being unset in the `Simplify.UpdateStateAndRemoveStyleAttributes()` method shortly thereafter, so it was absent from footnotes in the output. This was because the namespace was not being explicitly set, which is now fixed.
 - Created an abstract `LineEnricher` class to allow for the creation of new enrichers without significant code duplication. Adjusted the existing `QuotedText` enricher accordingly.
 - Created the `FootnoteEnricher` class to remove redundant sets of parentheses from footnotes.
   - There is a quirk in the drafting of Footnotes in Lawmaker Statutory Instruments which causes redundant sets of parentheses—one around the footnote reference, and one inside the footnote itself (at the beginning). These required removal.
  - An example of the output without the enricher—
    > `<p>In accordance with section 127(<authorialNote marker="7" class="footnote"><p>()2008 c. 29. Section 127 was amended by section 23(2)(a), 23(2)(b) and 23(2)(c) of the Growth and Infrastructure Act 2013 (c. 27) and by paragraph 64(2) of Schedule 13(1) to the Localism Act 2011 (c. 20).</p></authorialNote>) of the 2008 Act, the Secretary of State has applied the relevant tests and is satisfied that they have been met.</p>`


jurisdatum (approved):
Looks great guys! Sorry it took me so long to review

PR #169, opened 2025-08-27
LNI-315: Handle conclusions

erciyass:


mcpheed1:
Tested the following—
  - Explanatory Note markup is correct, with and without subheadings.
  - Commencement History Table markup is correct, with and without subheadings.
  - BlockLists are marked up as such.
  - Correctly handled different combinations of tblocks.
  - Added test coverage
  - Tests pass

LGTM!

PR #179, opened 2025-09-18
Change URI format for EMs

QR-TNA:
Within the metadata change URI suffix from /em to /memorandum for explanatory memoranda, data.xml to data.akn and rerun test files through parser


QR-TNA:
> This looks great. You've made consistence changes both to the main document and annex metadata, and you've updated the ExtractShortURIComponent method. Well done.
> 
> I'm wondering: if we change the URI to `.../data.akn`, should be change the FRBRFormat to `application/akn+xml`?

Thanks for this suggestion. This is now implemented and the test files have been updated to reflect this.

PR #186, opened 2025-09-24
LNI-336

mcpheed1:
**_LNI-336
Import Welsh Statutory instruments_**
The Senedd (Welsh Parliament) drafts legislation in both the English _and_ Welsh languages. The strict purpose of this ticket was simply to add Welsh counterparts to all pre-existing English string patterns. However, I spent some extra time creating a system which should hopefully be easy to extend to support any arbitrary number of languages in future.

Some specifics -
- The parser now accepts a list of ISO 639-3 language strings as input. These are the 'active' languages with which to parse the document.
- The parser initialises an instance of the new  `LanguageService` class with this list of languages.
- Throughout the codebase, rather than specifying a _single_ English string for each pattern-matching operation, we now instead specify a _map_ of strings (with one entry for each supported language).
- The `LanguageService.IsMatch()` method is used to pattern-match with all 'active' languages. 
  - For example, if `LanguageService` is initialised with ENG and CYM, then `LanguageService.IsMatch()` returns `true` if either the ENG _or_ CYM pattern is successfully matched.


erciyass:
LGTM!

PR #200, opened 2025-10-03
Sprint 36 merge

mcpheed1:


src/lawmaker/parsers/Conclusions.cs
```diff
@@ -0,0 +1,220 @@
...
+                block = Document.Body[i].Block;
+
+                if (block is WLine line && IsCenterAligned(line))
+                    break;
+
+                if (Match(LdappTableBlock.Parse) is LdappTableBlock tableBlock)
+                    blocks.Add(tableBlock);
+                else
+                    blocks.Add(block);
+                i += 1;
```
jurisdatum, line 145: Should this `i += 1;` be _inside_ the else block, like it is on line 65?

src/lawmaker/util/LanguageService.cs
```diff
@@ -0,0 +1,77 @@
...
+            this.languages.Add(Lang.ENG);
+    }
+
+    /// <summary>
+    /// Initializes a new instance of LanguageService with the specified languages
+    /// (represented as ISO 639-3 language strings). Defaults to English if no languages are provided,
+    /// </summary>
+    /// <param name="languages">Collection of ISO 639-3 language codes as strings</param>
+    public LanguageService(IEnumerable<string> languages)
+        : this(ParseLanguageStrings(languages))
```
jurisdatum, line 43: Might we need to guard against a null value here? I'm not sure whether the parameter might ever be null in Program (the CLI) or the Lambda function.

src/Program.cs
```diff
@@ -32,8 +32,9 @@ static Program() {
             new Option<bool>("--test", description: "whether to test the result"),
             new Option<FileInfo>("--attachment", description: "an associated file to include") { ArgumentHelpName = "file" },
             new Option<string>("--hint", description: "the type of document: 'em' or a Lawmaker type such as 'nipubb', 'uksi', or 'ukprib'"),
-            new Option<string>("--subtype", description: "the subtype of document e.g. 'order'"),
-            new Option<string>("--procedure", description: "only applicable --hint is a secondary type - the subtype of document e.g. 'order'"),
+            new Option<string>("--subtype", description: "the subtype of the document e.g. 'order'. Only applicable if --hint is a secondary type"),
+            new Option<string>("--procedure", description: "the proceedure of the document - the subtype of document e.g. 'order'"),
```
jurisdatum, line 36: Small typo in the word "proceedure". Also, it might be worth distinguishing this option more clearly from `--subtype`

jurisdatum (changes requested):

src/Program.cs
mcpheed1: Agreed, seems like a botched copy-paste operation. Have fixed the typos and rewritten the description.

src/lawmaker/util/LanguageService.cs
mcpheed1: Agreed! Null-safety added.

src/lawmaker/parsers/Conclusions.cs
mcpheed1: Yep, good catch. This would have caused the line immediately following the Commencement History table to be skipped over, and hence be absent from the parsed output.

mcpheed1:
Thanks @jurisdatum and @anthonyhashemi for taking the time to review the changes! I've since addressed each of the comments.

PR #392, opened 2026-04-30
Backlog: Make case number optional, support multiple case numbers and allow semicolons as list separators

dynamictulip:
Updated the CSV backlog processing to support multiple case numbers per entry and made the `CaseNo` column optional.
Also now support both commas and semicolons as separators for multi-value fields.

### Changes

- **Adjust boolean skip converter registration**: Make it clear that the way the boolean skip converter is registered is a one off
- **Shared csv array conversion code**: Moved jurisdiction array splitting into a converter called `DelimitedArrayConverter`
- **Field Type Update**: Changed `CsvLine.CaseNo` from `string` to `string[]` and applied the `DelimitedArrayConverter`.
- **Optional Column**: Made `CaseNo` optional in the CSV schema
- **Multi-Separator Support**: Updated `DelimitedArrayConverter` to split input strings by both `,` and `;`


backlog/src/Csv/DelimitedArrayConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+using System.Linq;
+
+using CsvHelper;
+using CsvHelper.Configuration;
+using CsvHelper.TypeConversion;
+
+namespace Backlog.Csv;
+
+internal class DelimitedArrayConverter : DefaultTypeConverter
```
jacksonj04, line 12: Hooray for well-encapsulated concerns!

jacksonj04 (approved):

PR #402, opened 2026-05-11
Dockerfile improvements

dynamictulip:
Some optimisations to improve build speed and caching and to control what is in the published TRE parser docker image.

## Changes:

- Add .dockerignore to stop entire repo being copied to build stage
- Only copy publishable files to final Docker image (rather than the entire source code)
- Cache dotnet restore for individual projects
- Combine dnf upgrade and cleanup steps to optimise image size

### DIFF
--- a/TRE/DockerfileV2
+++ b/TRE/DockerfileV2
@@ -20,8 +20,10 @@ RUN dotnet publish TRE/TRE.csproj -c Release -o /src --no-restore
 
 FROM base AS final
 
-RUN dnf upgrade -y
-RUN dnf clean all --enablerepo=\*
+# Use one RUN so package metadata does not clog image layer
+RUN dnf upgrade -y && \
+    dnf clean all --enablerepo=\* && \
+    rm -rf /var/cache/dnf
 
 WORKDIR /var/task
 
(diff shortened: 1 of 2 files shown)


TRE/DockerfileV2
```diff
@@ -4,21 +4,29 @@ WORKDIR /var/task
...
 
 WORKDIR /src
-COPY ["TRE/TRE.csproj",""]
-COPY . ./
+# Copy each csproj over then restore so this expensive operation can be cached
+COPY ["NationalArchives.FindCaseLaw.Utils/NationalArchives.FindCaseLaw.Utils.csproj", "NationalArchives.FindCaseLaw.Utils/NationalArchives.FindCaseLaw.Utils.csproj"]
+RUN dotnet restore NationalArchives.FindCaseLaw.Utils/NationalArchives.FindCaseLaw.Utils.csproj
+COPY ["judgments.csproj", "judgments.csproj"]
+RUN dotnet restore judgments.csproj
+COPY ["TRE/TRE.csproj", "TRE/TRE.csproj"]
```
anthonyhashemi, line 12: is this an arbitrary ordering? Or is it most likely that the frequency of change ranks up from `FindCaseLaw.Utils`<`judgments.csproj` < `TRE/TRE.csproj` ?

dynamictulip: In this case, it's a dependency chain - `TRE/TRE.csproj` depends on `judgments.csproj` which depends on `FindCaseLaw.Utils`
If we were to do it in an alternative order then the dotnet restore would fail

anthonyhashemi: I see!

anthonyhashemi (approved):

PR #405, opened 2026-05-12
Backlog: Split DI container configuration from DI container creation

dynamictulip:
Configuration and application setup in the Backlog parser is making it harder than necessary to make refactoring changes. This is a step towards using an apphost and more flexible configuration set up

## Changes

- Clearly define root namespace in csproj to stop linter warnings about capitalisation
- Split up DI configuration so it doesn't create a service provider
- Rework `ConfigureDependencyInjection_RegistersSystemTimeProvider` test


jacksonj04 (approved):
🚀

PR #427, opened 2026-05-18
Fix parser numbering bug

dynamictulip:
A document failed parsing because the list level format was slightly different to expected cases. Upon inspection it appears that there was a great deal of duplicated code and unnecessary complexities in the area that deals with transforming number formats and following this convention would lead to a greater explosion of regexes.

This PR cleans up that area of the codebase and fixes the issue but I was unable to introduce unit tests into the area without significant refactoring of much larger swathes of the codebase.

## Changes

- Unused and obsolete code removed from `Numbering2`
- Many regexes consolidated into a single transform
- `Magic2` renamed to `GetLevelIndicatorForParagraph` and split up


src/docx/Numbering2.cs
```diff
@@ -121,102 +121,152 @@ public static bool HasEffectiveStyleNumber(Paragraph paragraph) {
...
 
         /* None */
-        if (baseLevel.NumberingFormat.Val == NumberFormatValues.None) { // EWHC/QB/2009/406
-            if (string.IsNullOrEmpty(format.Val.Value))
+        if (numberingFormatVal == NumberFormatValues.None) { // EWHC/QB/2009/406
+            if (string.IsNullOrEmpty(docxLevelTextFormat))
                 return "";
-            logger.LogDebug("None number format: " + format.Val.Value);
-            return Regex.Replace(format.Val.Value, @"%\d", "");  // EWHC/Ch/2014/4092, [2023] EWHC 1526 (Admin)
+            logger.LogDebug("None number format: {LevelFormat}", docxLevelTextFormat);
```
dragon-dxw, line 137: Should this be `Non-number` whilst we're here?

src/docx/Numbering2.cs
```diff
@@ -121,102 +121,152 @@ public static bool HasEffectiveStyleNumber(Paragraph paragraph) {
...
-                return format.Val.Value;
-            // if (format.Val.Value == "%1")  // 00223_ukut_iac_2015_mk_sierra leone
-
-            if (format.Val.Value == Char.ConvertFromUtf32(0xf020)) { // [2024] EWHC 2427 (Ch)
-                logger.LogWarning("removing bullet xf020");
-                return null;
+        if (numberingFormatVal == NumberFormatValues.Bullet)
+        {
+            if (baseLevel.NumberingSymbolRunProperties?.RunFonts?.Ascii?.Value is not null &&
+                baseLevel.NumberingSymbolRunProperties.RunFonts.Ascii.Value.StartsWith("Wingdings"))
```
dragon-dxw, line 145: Do we feel there are sufficient test-cases for the complex logic? (I assume it's all "works against real files")

src/docx/Numbering2.cs
```diff
@@ -121,102 +121,152 @@ public static bool HasEffectiveStyleNumber(Paragraph paragraph) {
...
             m => TransformLevelNumber(m, main, paragraph, numberingId));
 
         return result;
     }
 
+    private static string TransformBullet(string docxLevelTextFormat)
+    {
+        switch (docxLevelTextFormat)
+        {
+            case "-": // EWHC/QB/2018/2066
```
dragon-dxw, line 180: I feel like there's additional refactoring that's worth doing here. I think there's about four different ways of saying "if the bullet is *this* then the bullet should be *this*" here. ("Yes, but not now" is fair, though)

dragon-dxw (approved):

src/docx/Numbering2.cs
dynamictulip: Afraid it's all "works against real files". 
I did try getting unit tests in around this area but rolled it all back after a time box because it was clear that it was going to take much longer. This part of the application is closely coupled to itself and to the openxml libraries as well as being filled with statics which makes it extremely difficult to mock anything out.

src/docx/Numbering2.cs
dynamictulip: I know what you mean - unfortunately I couldn't bring it all into a single switch statement because only compile-time constants can be used in a switch. I'd like to think it's cleaner than it was before though with at least some of the ifs removed!

src/docx/Numbering2.cs
dynamictulip: I don't think so - it's still a level format, just a `None` type rather than a `Bullet` or any of the number types

PR #439, opened 2026-05-21
FCLC-1938: Fix parser image uri

dynamictulip:
Fixes issue with getting images from the footer and endnotes and also adds first docx unit tests

## Changes

- Make docx relationships code all look at footer and end notes as well as header and body
- Add unit tests for docx parsing

### DIFF
--- a/src/docx/Relationships.cs
+++ b/src/docx/Relationships.cs
@@ -1,6 +1,6 @@
+#nullable enable
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 
 using DocumentFormat.OpenXml;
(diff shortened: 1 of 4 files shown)


src/docx/Relationships.cs
```diff
@@ -9,67 +9,46 @@
...
     }
 
-    public static Uri GetUriForImage(HeaderPart header, StringValue relationshipId) {
-        OpenXmlPart part = header.Parts.Where(part => part.RelationshipId == relationshipId.Value).First().OpenXmlPart;
-        return part.Uri;
+    private static OpenXmlPart GetMainPartFromElement(OpenXmlElement element)
+    {
+        var root = element.Ancestors<OpenXmlPartRootElement>().Single();
+
+        return root switch
```
anthonyhashemi, line 38: cant find docs on 
`return varialble switch...` can you explain or link to relevant doc 🙏

test/Docx/TestRelationships.cs
```diff
@@ -0,0 +1,79 @@
...
+        where T : OpenXmlPart, ISupportedRelationship<ImagePart>
+    {
+        using var mockWordprocessingDocument = new MockWordprocessingDocument();
+
+        var (rootElement, part) = getDocumentArea(mockWordprocessingDocument);
+        var uri = new Uri("https://example.com/external-image.png");
+
+        var drawing = rootElement.AppendChild(new Paragraph())
+                                 .AddExternalImageToDocument(part, uri, out var relationshipId);
+        var relId = new StringValue(relationshipId);
```
anthonyhashemi, line 75: where is `relationshipId` defined/set?

test/Mocks/MockWordprocessingDocument.cs
```diff
@@ -0,0 +1,175 @@
...
+using DocumentFormat.OpenXml.Packaging;
+using DocumentFormat.OpenXml.Wordprocessing;
+
+using Drawing = DocumentFormat.OpenXml.Drawing;
+using DrawingWordProcessing = DocumentFormat.OpenXml.Drawing.Wordprocessing;
+using DrawingPictures = DocumentFormat.OpenXml.Drawing.Pictures;
+
+namespace test.Mocks;
+
+public sealed class MockWordprocessingDocument : IDisposable
```
anthonyhashemi, line 16: This is a very impressive and intricate test fixture creator for documents (it is in fact creating a fully compliant word doc, rather thab mocking anything, right?)! Is there much benefit in using and maintaining this vs creating simple word docs we add to the repo which we use as test docs for the scenarios it is being used for?

test/Mocks/MockWordprocessingDocument.cs
```diff
@@ -0,0 +1,175 @@
...
+            new Run(new FootnoteReference())));
+    }
+
+    public void Dispose()
+    {
+        wordDocument.Dispose();
+    }
+}
+
+public static class MockWordprocessingDocumentExtensions
```
anthonyhashemi, line 66: this isn't referenced anywhere - or is the suffice Extensions a magic naming convention thing?

src/docx/Relationships.cs
dynamictulip: In C# there are two types of `switch`:
- `switch` statement which is a traditional `if`/`else` stand that can do multiple operations when a case is matched - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/selection-statements#the-switch-statement
- `switch` expression which has cleaner syntax and can use advanced pattern matching to calculate and return an object - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression

This is a [C# switch expression](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression).

test/Docx/TestRelationships.cs
dynamictulip: This is an `out` parameter which allows a method to return multiple things - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/method-parameters#out-parameter-modifier
It's defined on line 74 `out var relationshipId` and set inside `AddExternalImageToDocument`.
In all honesty, there's no particular reason to use this over a tuple return other than it being easier to use my refactoring tools on during development. If it is confusing then I can remove it

test/Mocks/MockWordprocessingDocument.cs
dynamictulip: This test setup is indeed intricate and daunting, however so is the code that it's testing and it's important to have an understanding of the structure of docx files when making changes to that prod code. You're right that it's not really a mock - I started down that route and decided that it wasn't worth pursuing too far.

The biggest benefit of using this over simple sample test docs is that it can get right into the nooks and crannies of how the docs are configured under the hood and thus explicitly replicate the kinds of edge case scenarios that we need to cope with for any given unit test.

I'd be cautious about relying entirely on sample docs as it would be extremely easy to accidentally lose some hidden link or formatting we were testing for without noticing and then to gain false confidence from the tests.

On the other hand, it's perfectly possible to incorrectly set up the test scenarios in an entirely coded situation and get false positives from the tests that way! When I wrote these sample docs to file and opened them in Pages on mac, I discovered that I had initially put them together wrong and nothing was visible!

I think the best approach would probably be blended - we could have a few different starting sample docs (with different underlying configurations as needed) that we then add extra coded setup to for any intricate edge case tests. In order to defend against false positives we could enable [mutation testing](https://stryker-mutator.io/docs/stryker-net/introduction/) across the docx tests and put a PR gate to ensure that the mutation score doesn't drop. All of this is probably a ticket or two in its own right though

test/Mocks/MockWordprocessingDocument.cs
dynamictulip: There's no magic naming conventions in C# (that I can think of) - these are extension methods which add extra capabilities to pre-existing classes via the `this` keyword on the first parameter of a method - https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-implement-and-call-a-custom-extension-method#extension-methods-earlier-versions
We are on C# 12 / .NET 8 but in C# 14 / .NET 10 there's a new `extension` keyword to make this more explicit.

test/Mocks/MockWordprocessingDocument.cs
anthonyhashemi: Yeah I quite like that blended approach. Also we could add assertions to the test fixtures themselves (or at the start of the test to ensure our starting point is as expected).

dynamictulip: So good to merge now and revise in future?

test/Docx/TestRelationships.cs
anthonyhashemi: will pick your brains on slack

test/Mocks/MockWordprocessingDocument.cs
anthonyhashemi: I meant that the suffix `Extensions` on the class seems like a magic glue - again will chat on slack

src/docx/Relationships.cs
anthonyhashemi: right, root is the input to the swtich

anthonyhashemi (approved):

PR #445, opened 2026-05-22
FCLC 1841 - Add bulk upload playbook

dynamictulip:
The Bulk upload process is triggered by the backlog parser but runs through multiple services - this PR adds documentation on how the wider process hangs together and provides tools to analyse different logs to determine the status of any given document.

## Changes

- Adds documentation on the wider bulk upload process and how to spot problems
- Adds scripts to gather and consolidate logs from MarkLogic and CloudWatch with the Tracker csv

### DIFF
--- a/backlog/docs/bulk-upload-process.md
+++ b/backlog/docs/bulk-upload-process.md
@@ -1,0 +1,162 @@
+# Wider Bulk Upload process
+
+<!-- TOC -->
+* [Wider Bulk Upload process](#wider-bulk-upload-process)
+  * [Bulk ingestion events](#bulk-ingestion-events-)
+  * [Running a bulk upload](#running-a-bulk-upload)
+  * [Verifying document status after a bulk upload](#verifying-document-status-after-a-bulk-upload)
+    * [Get logs as csv](#get-logs-as-csv)
+      * [Backlog tracker csv](#backlog-tracker-csv)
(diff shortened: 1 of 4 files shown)


backlog/docs/marklogic-ingested-parser-run.xquery
```diff
@@ -0,0 +1,67 @@
+xquery version "1.0-ml";
+import module namespace dls = "http://marklogic.com/xdmp/dls" at "/MarkLogic/dls.xqy";
+
```
dragon-dxw, line 3: Can we add a comment here about the overall goals of this file, when you should be running it?

backlog/docs/bulk-upload-process.md
```diff
@@ -0,0 +1,162 @@
...
+```mermaid
+graph TD
+    each[[For each document]] --> statePublished
+    statePublished{Is state Published?}
+    statePublished -- yes --> stateParserFailed
+    statePublished -- no --> ok([Not a concern])
+    stateParserFailed{Is state ParserFailed?}
+    stateParserFailed -- yes --> log(["
+    Document not sent to S3 so just
+    note parser failure reasons for
```
dragon-dxw: I don't think this sentence quite works.

Document not sent to S3. Record reasons for parser failure and fix later"

dragon-dxw (approved):

PR #448, opened 2026-05-26
Lawmaker SC Sprint 13

mcpheed1:
---
**_LDAPP-4857_
Handle images**

 - Images embedded in the body of a Word document are detected during import
 - Appropriate image markup is generated and included in the XML output
```
<tblock class="image centre">
    <p><img src="" alt="" width="" height=""/></p>
</tblock>
```
 - All images in the Word document are exported into the `--output-zip` archive alongside the generated XML.
 - Image references in the XML correctly correspond to the image files in the zip archive
---
**_SPT-1927_
Fix infinite loop issue caused by blank lines after block lists**

- The `while` loop at line 225 in `BlockList.Parse()` was missing a break statement, causing the parser to infinitely reprocess the same blank line without advancing. This resulted in the application hanging during document parsing.
---
**_LSC-189_
Add catch-all handler for unhandled LISTNUM field codes**

Adds a general-purpose fallback method to handle any LISTNUM field codes not explicitly covered by existing logic. The primary goal is to prevent exceptions being thrown due to unrecognised field codes during Word document parsing. The numbering output from this handler may not be perfectly accurate in all cases, but it ensures the import process completes without failure.

Also: 

- Fix bug preventing a heading-less Prov1 from before recognised when definition children are involved
- Add handling for more types when using GetLines
- Add a test that involves all the described cases
---
**_LSC-181_
Handle empty quoted structures**

- Word documents can contain quoted structures with a single provision that contains only a number and no actual content. When the parser encounters these empty provisions, it results in a null `IInline` reference, causing an unhandled exception.
- Added a null check for `IInline` to gracefully handle this scenario.

---
**_LSC-182_
Miscellaneous prioritised fixes**

This ticket handles a collection of small fixes which were identified by TNA as high priority.

- Removed requirement that Schedule Cross Headings be centre-aligned, to support UK, SP, and SC bills, in which they are left-aligned.
- Made the em-dash between SchProv1 and their 1st SchProv2 child optional, as it is only present in NI & SI.
image
- No longer require Table Numbers or Captions to be left-aligned. Instead, allow ANY alignment, and reflect this alignment in class attribute of the element in the XML (i.e. class="centre").
- Allow Prov2 elements to have (italic) headings.
- Prevent Para1 elements which happen to have numbers that are roman numerals (i, l, v, x) from being incorrectly parsed as Para2 by sequentially comparing the number of the current and previous paragraphs during parsing.
image

---


jurisdatum:
@mcpheed1 Sorry I'm not done with this yet. I'm most of the way through, and should have comments for you by tomorrow

src/lawmaker/models/Prov.cs
```diff
@@ -128,11 +128,22 @@ public static bool IsValidNumber(string num)
...
+        /// <c>true</c> if the text begins with a parenthesized alphanumeric identifier (e.g., "(1)", "(A1)", "(AB12)"),
+        /// optionally after whitespace and/or an em dash.
+        /// </returns>
+        /// <remarks>
+        /// The em-dash is present in NI and SI jurisdictions and absent in UK, SP and SC jurisdictions.
+        /// </remarks>
         public static bool IsFirstProv2Start(string text)
         {
-            string pattern = @"^\s*\u2014\([A-Z]*\d+(?:[A-Z]+\d+)*[A-Z]*\).*";
+            string pattern = @"^\s*\u2014?\([A-Z]*\d+(?:[A-Z]+\d+)*[A-Z]*\).*";
```
jurisdatum, line 146: This widens the accepted first-Prov2 prefix to include no em dash and alphanumeric nums like `(A1)`/`(AB12)`, but `FixFirstProv2` still assumes the old fixed shape: it emits `new("(1)", ...)` and removes `Substring(5)` in both inline-split paths. That corrupts newly accepted inputs: `(1) text` drops the first body character, and `(A1) text`/`—(A1) text` emits the wrong `<num>`. Could we derive the captured num and remainder offset from this regex match instead of duplicating the old assumptions?

src/parsers/common/fields/ListNum.cs
```diff
@@ -115,8 +117,66 @@ internal static INumber Parse(MainDocumentPart main, Run first, string fieldCode
...
+    private static string CatchAllListNums(MainDocumentPart main, Run first, string fieldCode) {
+        string fieldName = null;
+        Match match = Regex.Match(fieldCode, @"^ LISTNUM (""?([^""\\]+?)""?) ");
+        if (match.Success) {
+            fieldName = match.Groups[1].Value;
+        }
+        int fieldLvl = -1;
+        match = Regex.Match(fieldCode, @"\\l (\d)");
+        if (match.Success) {
+            fieldLvl = int.Parse(match.Groups[1].Value) - 1; // fieldLvl is 0 based
```
jurisdatum, line 147: `fieldLvl` is converted to 0-based here, which is what `FormatNumber` expects below, but the no-`\s` branch also passes it to `FormatNNumberDefault`. I believe that that formatter expects the Word 1-based level (`l == 1..9`), as the explicit handlers above pass `\l` through unchanged. So `LISTNUM ... \l 1` with no `\s` calls `FormatNNumberDefault(0, ...)` and throws, and `\l 2+` formats one level too shallow. Could we keep separate locals for the 0-based `ilvl` and the 1-based `level`, or pass `fieldLvl + 1` into `FormatNNumberDefault`?

src/lawmaker/parsers/Prov2.cs
```diff
@@ -25,16 +31,36 @@ private HContainer ParseProv2(WLine line)
...
+        /// If these conditions are met, the parser advances the cursor and returns the heading line.
+        /// Otherwise, no state is changed and null is returned.
+        /// </remarks>
+        private WLine? ConsumeProv2Heading(WLine line)
+        {
+            if (line is WOldNumberedParagraph)
+                return null;
+            if (!line.IsAllItalicized())
+                return null;
+            if (IsAtEnd())
```
jurisdatum: This checks the current cursor, but I think the risky access is after `i += 1`: when the italic line is the final block, `i` advances to `Body.Count`, then `ParseProv2` reads `Body[i]` at line 19 and throws `ArgumentOutOfRangeException`. The guard needs to prove there is a following block to parse as the numbered Prov2. Could we use `if (i + 1 >= Body.Count) return null;` here before advancing? Memoization can restore `i` on a null return, but not after this throw.

src/lawmaker/akn/Builder.cs
```diff
@@ -463,6 +490,30 @@ protected override void AddFootnote(XmlElement parent, IFootnote fn)
...
+                    style.Split(":")[0], 
+                    PtToMM(float.Parse(Regex.Replace(style, @"[^0-9.-]", ""))).ToString());
+            }
+        }
+
+        protected override void AddImageRef(XmlElement parent, IImageRef model) {
+            XmlElement img = doc.CreateElement("img", ns);
+            img.SetAttribute("src", "/document/image?filename="+model.Src+"&ds=LEGI_DRAFTING");
+            img.SetAttribute("alt", model.Src);
+            ExtractDimensions(img, model);
```
jurisdatum: This calls `ExtractDimensions` before the null guard below, but `ExtractDimensions` immediately does `model.Style.Split(";")`, so an image ref with no inline style will throw before it can be emitted. Could we move the `model.Style is not null` check above this call, or make `ExtractDimensions` a no-op for null/empty styles? While touching this, `ExtractDimensions` should also skip non-numeric style fragments or use `TryParse`; empty fragments/trailing semicolons and declarations like `display:block` currently reach `float.Parse("")` and throw.

src/lawmaker/akn/Builder.cs
```diff
@@ -463,6 +490,30 @@ protected override void AddFootnote(XmlElement parent, IFootnote fn)
             IEnumerable<IBlock> content = FootnoteEnricher.EnrichInside(fn.Content);
             blocks(authorialNote, content);
         }
+        private double PtToMM(float pt)
+        {
+            return Math.Floor(pt * 25.4/300*1.44); //transforming between different measurements
```
jurisdatum, line 500: Can we clarify the unit conversion here? `25.4 / 300 * 1.44` is about `0.122` per input unit, while standard points-to-mm would be `25.4 / 72` (`0.353`). The method name and `pt` parameter suggest points, but the factor looks calibrated for some other source/target unit. If this is intentional, a short comment naming the input unit and expected output unit would make this much safer to maintain.

src/lawmaker/akn/Builder.cs
```diff
@@ -388,7 +413,9 @@ private ILine TrimLine(ILine line)
                 .SkipWhile(IInline.IsEmpty).Reverse();
 
             // Trim start of first inline
-            IInline first = newContents.First();
+            IInline first = null;
+            if (newContents.Any())
```
jurisdatum: For an all-whitespace line, this trimming chain leaves `newContents` empty even though `line.Contents.Count() != 0`. In that case `first` remains `null`, and the later `trimmedInlines.Add(first)` adds that null into the returned line's contents. That looks like why `Inline.cs` needed the new `model is null` branch. Could we handle the empty case here instead, e.g. return early when `newContents` is empty or guard the later add with `first is not null`, so null never enters the inline list?

src/lawmaker/akn/Builder.cs
```diff
@@ -365,7 +369,28 @@ override protected void p(XmlElement parent, ILine line) {
...
+                block.SetAttribute("style", CSS.SerializeInline(styles));
+            ContainingParagraphStyle = stripped.Style;
+            foreach (IInline inline in stripped.Contents)
+                // If line contains image, wrap with tblock first
+                if (inline is WImageRef wimageRef)
+                {
+                    XmlElement tblock = doc.CreateElement("tblock", ns);
+                    parent.AppendChild(tblock);
+                    tblock.SetAttribute("class", ns, "image centre");
+                    tblock.AppendChild(block);
```
jurisdatum, line 388: `AppendChild` moves an existing node, and this runs once per image inline. For a line with two images, the first image moves `block` into `tblock1`; the second creates `tblock2` and moves the same `block` out of `tblock1` into `tblock2`, leaving `tblock1` empty. The single-image fixture passes, but multi-image lines produce an empty wrapper plus one wrapper containing the whole line. Could we lift the wrapping out of the loop: create one `tblock` when the line contains any image, move `block` into it once, then add all inlines normally?

src/lawmaker/parsers/Table.cs
```diff
@@ -102,10 +100,10 @@ partial record LdappTableNumber(
     List<WLine>? Captions
 ) : ILineable {
 
-    private static readonly LanguagePatterns TableNumberPatterns = new()
+    internal static readonly LanguagePatterns TableNumberPatterns = new()
     {
-        [Lang.EN] = [@"^Table\s+\w+$"],
-        [Lang.CY] = [@"^Tabl\s+\w+$"]
+        [Lang.EN] = [@"^(?:The\s+Table|Table\s+\w+)$"], // e.g. matches The Table & Table 1
```
jurisdatum, line 105: Adding the bare `The Table` / `Y Tabl` alternatives makes any paragraph with exactly that text match `LdappTableNumber`. That has two knock-on effects: if a real table appears later, `LdappTableCaptions` (`AdvanceWhile(block => block is not WTable)`) can swallow every intervening paragraph as captions; and in `BreakFromProv1`, a centre-aligned `The Table` now satisfies `Peek(LdappTableNumber.Parse)`, so it no longer terminates the enclosing Prov1. Could we gate these numberless forms more tightly, e.g. only accept them when the table follows immediately / after known caption lines, or have `BreakFromProv1` peek the full table block rather than the number line alone?

src/lawmaker/parsers/ScheduleCrossHeading.cs
```diff
@@ -62,8 +62,8 @@ private bool PeekScheduleCrossHeading(WLine line)
             }
             else
             {
-                if (!line.IsCenterAligned())
-                    return false;
+                // Left-aligned in UK, SP & SC Bills, center-aligned in NI Bills.
+                // Best to not use alignment at all
```
jurisdatum: Removing the centre-alignment check leaves `IsPartiallyItalicized()` as the only discriminator here. That seems broad for schedule body text: an unnumbered paragraph with an italic phrase can become a cross-heading if provisions follow. Should we add another structural cue, or make the alignment rule jurisdiction-aware, so ordinary partially italic prose isn't claimed as a heading?

src/parsers/common/fields/ListNum.cs
```diff
@@ -115,8 +117,66 @@ internal static INumber Parse(MainDocumentPart main, Run first, string fieldCode
...
+        }
+        int fieldStart = -1;
+        match = Regex.Match(fieldCode, @"\\s (\d)");
```
jurisdatum: `(\d)` only captures one digit, so multi-digit starts are truncated: `\s 12` and `\s 10` are both read as start `1`. `fieldStart` then feeds the emitted number below, so explicit starts >= 10 produce silently wrong list numbers. Could this use `\d+` and parse the full switch value? The `\l` regex above has the same single-digit shape, though Word levels are normally <= 9.

src/parsers/common/fields/ListNum.cs
```diff
@@ -115,8 +117,66 @@ internal static INumber Parse(MainDocumentPart main, Run first, string fieldCode
...
+    /*
+     * This method is a general solution to handle any LISTNUM field codes not explicitly handled.
+     * The fieldNum returned is expected to not be correct in most cases as the main purpose of this 
+     * method is to prevent exceptions being thrown due to unhandled field codes.
+     */
+    private static string CatchAllListNums(MainDocumentPart main, Run first, string fieldCode) {
+        string fieldName = null;
+        Match match = Regex.Match(fieldCode, @"^ LISTNUM (""?([^""\\]+?)""?) ");
+        if (match.Success) {
+            fieldName = match.Groups[1].Value;
```
jurisdatum: Group 1 is the outer capture (`"?...?"`), so it includes the optional surrounding quotes; group 2 is the bare name. Reading group 1 means a quoted name like `"WP List 1"` can be passed to `FormatNumber` with the quotes included, where the abstract-numbering lookup won't find it and can throw. Simple quoted forms are handled by the explicit patterns above, so this affects quoted named-list variants that fall through to the catch-all. Could this read `match.Groups[2].Value` instead?

src/Program.cs
```diff
@@ -84,18 +87,24 @@ static int Transform(ParseResult parseResult)
...
+        LegislationClassifier classifier = new LegislationClassifier((DocName) docName, subType, procedure);
+        LanguageService languageService = new LanguageService(language);
+        Response response = Helper.LocalParse(input.FullName, classifier, languageService);
+        if (output is not null)
+            File.WriteAllText(output.FullName, response.Xml);
+        else
+            Console.WriteLine(response.Xml);
+
+        if (outputZip is not null)
+            SaveImagesToZip(response.Images, outputZip);
```
jurisdatum: Just checking the intended CLI contract here: this zip now contains images only, whereas the previous `SaveZip(response, outputZip)` path wrote `judgment.xml`, `meta.json`, and images into the archive. The PR description says images are exported “alongside the generated XML”; is the intended shape now separate XML output plus an image-only zip?

jurisdatum (commented):
Thanks for putting this together. It’s a substantial PR, and the new XML docstrings plus the `IsSubsequent*` renames make this part of the parser noticeably easier to follow.

I’ve left a number of line comments. I’m reviewing this from outside the day-to-day context of the project. so I may be missing assumptions, fixtures, or real-world document patterns that informed these changes. Please treat the comments as things to check rather than firm conclusions. You’re much better placed than I am to judge which ones are real issues and which are acceptable tradeoffs.

mcpheed1:
Thanks @jurisdatum, very much appreciate your feedback as always. We'll be sure to address your review comments as soon as possible. 

As a sidenote: we originally batched multiple tickets into each PR to avoid flooding you with constant review requests. Now that our development pace on the judgment parser has slowed, it probably makes more sense to submit PRs on a per-ticket basis. They'd be smaller, easier to review, and easier to reason about. Would you agree?

jurisdatum:
> As a sidenote: we originally batched multiple tickets into each PR to avoid flooding you with constant review requests. Now that our development pace on the judgment parser has slowed, it probably makes more sense to submit PRs on a per-ticket basis. They'd be smaller, easier to review, and easier to reason about. Would you agree?

Sure, I can handle more, smaller PRs. Whatever's best for you

erciyass:
Hi @jurisdatum, I've picked up the ticket that was created for addressing all of your comments and getting this PR merged so expect me to gradually reply to your comments as I go through them. Whether that's me deciding it's not an issue worth addressing, if I decide to make a change or if I have further questions

src/lawmaker/models/Prov.cs
erciyass: I believe it's currently okay to assume the prov2 num will always be (1). Seems to be the case with all of our test data so I'm going to mark this resolved as I don't think it's worth the effort to address

src/parsers/common/fields/ListNum.cs
erciyass: I'll be making the change to pass `fieldLvl + 1` into `FormatNNumberDefault`

src/lawmaker/parsers/Prov2.cs
erciyass: Addressed this by changing line 59 to check `if (i > Body.Count - 2)`

src/lawmaker/akn/Builder.cs
fangskev: This is an approximation to how the lawmaker editor converts image size into mm. However it's an imperfect approximation resulted from trial and error. It seems that the lawmaker editor does not use the standard conversion of `25.4/72` either. So this may warrant additional investigation at a later time.

src/lawmaker/akn/Builder.cs
erciyass: 6. I made a change for this returning early if newContents is empty

src/lawmaker/parsers/Table.cs
erciyass: Hi Jim, correct me if I'm wrong but my understanding is an issue may only arise if there happens to be a `p` which matches the regex (For example, the text is exactly `The Table`)?
If so, I believe this would be another case that is probably worth leaving alone in favour of saving effort because it would be an extremely unlikely edge case

src/lawmaker/parsers/ScheduleCrossHeading.cs
erciyass: 9. Addressed this by checking alignment based on jurisdiction

src/parsers/common/fields/ListNum.cs
erciyass: 10. Thanks for catching that. Addressed it by adding `+` after `\d`

src/parsers/common/fields/ListNum.cs
erciyass: I had assumed the quotes should be part of the name. Thanks for the clarification. Fix applied by using group 2 for `fieldName`

src/lawmaker/akn/Builder.cs
erciyass: 4. Fix applied with null check being moved up and only setting attributes which contain a digit in their value in `ExtractDimensions`

src/lawmaker/akn/Builder.cs
erciyass: I've added a comment above the method explaining what Kevin stated

src/lawmaker/akn/Builder.cs
erciyass: Fix applied. Only one tblock gets created now. Also updated a test to include this case

src/Program.cs
erciyass: Yes, what you said is correct.

src/lawmaker/models/Prov.cs
chatawayf: Hmm I think this might be a valid concern with the converted subsection num. If the first subsection is anything other than `(1)` then it is converted to `(1)` incorrectly.

I couldn't work out the case where the first body character gets dropped. In my testing:
* `(1) text` -> `<num>(1)</num><content>text</content>`
* `(1AAAA)` -> `<num>(1AAA</num><content>A)text</content>`
In `FixFirstProv2` since `text` is the first inline which I think is guaranteed to be the num so I don't think body characters will get dropped?

src/lawmaker/parsers/Table.cs
chatawayf: Just confirming this is known and expected behaviour. TNA are aware that a lone "Table 1" or similar has special significance and a table is expected shortly after it.

src/lawmaker/akn/Builder.cs
jurisdatum: I think this one is not quite fixed. Moving the guard up still allows `ExtractDimensions` to be called with a null style, and I believe a null style will throw _within_ `ExtractDimensions`. So maybe put the `ExtractDimensions` call within the `if (model.Style is not null)` block, or put a `if (model.Style is null) return;` check at the top of `ExtractDimensions`.

erciyass: Thanks for catching that, I've directly pushed a proper fix now!

jurisdatum (approved):
Thanks all — that's the last of my comments addressed. I've been back through the full set and everything checks out. Really nice work!

PR #452, opened 2026-05-27
FCLC-1950 Correctly handle primary source / input file mismatch in backlog parser

dragon-dxw:
If the primary source was a doc/rtf file, the parser recieves a docx, and must not rename it to a doc/rtf.
Equally, whilst doing this, we must not change the filename of the `primary_source` field.

We note there are issues with the filehash being incorrect at this time -- it's the hash of the docx, not the doc. 

[FCLC-1950](https://national-archives.atlassian.net/browse/FCLC-1950)

## Changes

- Append `.docx` to source file in tar.gz and in payload metadata for files parsed as `docx` which do not have `docx` file extension already in original file name
- Add test assertion for bundle file contents (and refactor around test zip files)

[FCLC-1950]: https://national-archives.atlassian.net/browse/FCLC-1950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


backlog/src/BacklogParserWorker.cs
```diff
@@ -269,12 +269,15 @@ private async Task<Bundle> GenerateBundleAsync(CsvLine csvLine, Guid parserRunId
 
         var externalMetadataFields = metadataTransformer.CsvLineToMetadataFields(csvLine);
 
-        var trePipelineMetadata = metadataTransformer.CreateFullTreMetadata(parserRunId, csvLine.FileName, mimeType, contentHash,
+        var was_word_file = new []{".rtf", ".doc", ".docx"}.Contains(Path.GetExtension(csvLine.FileName));
```
dynamictulip: There's a ton of different extensions that would need adding to this list (including typo extensions) so I think it'd be better to look at this from the point of view of `if not stub and not docx then append .docx to filename`.
That would also give us a clearer picture of where a file may have gone through conversion (because it'd be a `.doc.docx`)

test/backlog/ZipFileHelpers.cs
```diff
@@ -1,53 +1,51 @@
...
 public static class ZipFileHelpers
 {
-
-    public static List<string> GetListingFromZippedContent(byte[] content)
+    public static byte[] GetFileFromZippedContentAsBytes(this byte[] zipFileContent, string fileName)
     {
-        using MemoryStream memoryStream = new(content);
-        using GZipStream gz = new(memoryStream, CompressionMode.Decompress);
+        using MemoryStream zipFileMemoryStream = new(zipFileContent);
+        using GZipStream gz = new(zipFileMemoryStream, CompressionMode.Decompress);
```
dragon-dxw: At this point we've taken the bytestream of the tar.gz (`zipFileMemoryStream`) and decompressed it so what we actually have here is a `tarStream` rather than `gz`

dynamictulip (approved):

test/backlog/ZipFileHelpers.cs
dynamictulip: Done

PR #471, opened 2026-06-04
Backlog: Remove leading zeros from NCNs in external metadata

dynamictulip:
NCNs `[2025] UKUT 27 (LC)` and `[2025] UKUT 0027 (LC)` are the same and the main parser returns a cleaned version to the caller by trimming the leading zeroes from them as it extracts it from the document text. If the parser receives an override via "outside" metadata then it will always prefer that override to the document text, even if the NCN is in an invalid format or contains leading zeroes

Before now, the backlog parser was passing an unclean NCN to the main parser which was overriding the output and causing duplication detection issues downstream as the rest of Find Case Law doesn't expect leading zeroes in NCNs. This PR fixes this issue.

## Changes

- Trim leading zeroes in NCNs provided by external metadata (but keep the original record of what was read)
- Add E2E backlog tests with a doc containing an NCN


backlog/src/Csv/CsvLine.cs
```diff
@@ -81,6 +82,23 @@ internal record CsvLine
     [TypeConverter(typeof(BooleanSkipConverter))]
     public bool Skip { get; set; }
 
+    /// <summary>
+    /// NCN with zeros removed
```
jacksonj04, line 86: This is true at the moment, potentially in the future could do things like "the court code must always be in all caps". Not sure if it's worth expressing that "cleaned NCN" is really more of what we consider canonical, rather than just that it has no leading zeroes?

backlog/src/Csv/CsvLine.cs
```diff
@@ -81,6 +82,23 @@ internal record CsvLine
...
+    /// </summary>
+    [Ignore]
+    internal string? CleanedNcn
+    {
+        get
+        {
+            if (Ncn is null)
+                return null;
+                
+            return cleanedNcn ??= Regex.Replace(Ncn, " 0+", " ");
```
dragon-dxw, line 96: Yeah, this makes sense. The year is prefixed with `[`, the subcourt with `(` and the court can't start with a number.

dragon-dxw (approved):

backlog/src/Csv/CsvLine.cs
dynamictulip: Actually, that's probably something we want to implement now - the trigger for this is wanting to prevent multiple publications of cases because of variances in the external metadata NCN. Will tackle in a new PR

dynamictulip: Ah casing of court code property rather than court code within the NCN? The court code property has different validation within the parser and is a property we create for FCL in the external metadata.
Was about to make NCNs require upper courts then found "[2023] UKIPTrib 11" 🤦‍♀️

PR #490, opened 2026-06-11
Backlog: Enable dotnet user secrets

dynamictulip:
[dotnet user secrets](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?&tabs=linux%2Cpowershell#use-the-secret-manager-tool) is the preferred way in .NET to manage configuration that shouldn't be committed. This PR enables that capability and updates the documentation to describe how to use it as an alternative to environment variables.

## Changes

- Enables the use of dotnet user secrets

### DIFF
--- a/backlog/README.md
+++ b/backlog/README.md
@@ -17,7 +17,7 @@ This module provides a specialized entry point to the parser specifically design
       * [Optional Columns](#optional-columns)
       * [CSV Line Example](#csv-line-example)
     * [Tracker CSV](#tracker-csv)
-    * [Environment Variables](#environment-variables)
+    * [Configuration variables](#configuration-variables)
       * [AWS Configuration](#aws-configuration)
   * [Development](#development)
     * [Components](#components)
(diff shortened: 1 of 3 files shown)


backlog/README.md
```diff
@@ -160,18 +158,54 @@ This is created and populated by the Backlog Parser. It tracks which judgments h
...
+and can be set as follows:
+
+| Section       | Key                     | Description                                                                      |
+|---------------|-------------------------|----------------------------------------------------------------------------------|
+| BacklogParser | `CourtMetadataFilePath` | Path to the CSV file containing court metadata                                   |
+| BacklogParser | `DataFolderPath`        | Path to the folder containing judgment data files                                |
+| BacklogParser | `TrackerFilePath`       | Path to the CSV file tracking uploaded judgments                                 |
+| BacklogParser | `OutputFolderPath`      | Path to where generated bundle files will be saved                               |
+| BacklogParser | `BucketName`            | AWS bucket to upload processed files and xml to - optional if using dry run mode |
+| -             | `AWS_REGION`            | AWS region for S3 bucket operations - optional if using dry run mode             |
```
dragon-dxw, line 174: Not clear what the `-` means: would "project wide" or similar help?

backlog/README.md
```diff
@@ -160,18 +158,54 @@ This is created and populated by the Backlog Parser. It tracks which judgments h
...
+Using [.NET user secrets](https://learn.microsoft.com/en-us/aspnet/core/security/app-secrets?&tabs=linux%2Cpowershell#use-the-secret-manager-tool)
+for local development configuration
+    - Use `Section:Key` format for user secrets cli
+      ```bash 
+      dotnet user-secrets set "Section:Key" "value"
+      dotnet user-secrets list
+      ```
+    - Use JSON when editing the user secrets file via IDE
+      ```json
+      {
```
dragon-dxw, line 188: If this winds up in a file like `appsettings.json` is it worth having a `.appsettings.json.example` with sensible default values for the non-private variables?

backlog/README.md
dragon-dxw: It's literally "there is none and it's a bare string `AWS_...`"

dragon-dxw (approved):

PR #491, opened 2026-06-11
Backlog fixup env config

dynamictulip:
Tests were failing locally because the user secrets were overriding the environment variables set by the e2e tests. This PR fixes that by reinstating the original order of configuration precedence. It also adds in appsettings for quality of life logging defaults and removes the old third party .env file config as it's been replaced by user secrets

## Changes

- fix broken local tests suffering from env vs user secrets conflict
- remove dotenv file support
- add appsettings

### DIFF
--- a/backlog/src/Program.cs
+++ b/backlog/src/Program.cs
@@ -14,8 +14,6 @@ using Backlog.Src;
 using Backlog.Tracking;
 using Backlog.Utilities;
 
-using DotNetEnv.Configuration;
-
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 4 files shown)


backlog/src/Program.cs
```diff
@@ -170,10 +168,12 @@ private static int RunBacklogParser(bool isDryRun, uint? id, bool autoPublish)
     private static IHost CreateAppHost(bool isDryRun, uint? id, bool autoPublish)
     {
         var builder = Host.CreateApplicationBuilder();
-        builder.Configuration.AddDotNetEnv();
 
-        // explicitly add user secrets configuration provider so the Production dotnet environment can access it because we always run this application from local machines
+        // Explicitly add user secrets configuration provider so the Production dotnet environment can access it
+        // because we always run this application from local machines. Then readd the environment variables config
```
dragon-dxw: could we hyphenate `re-add` so it looks less like a typo of `read`

dragon-dxw (approved):

PR #501, opened 2026-06-17
test(leg): gate fixture regenerators behind UPDATE_XML

QR-TNA:
This PR stops the leg test-fixture regenerators from rewriting the golden .akn/.html fixtures on a normal test run, gating them behind the UPDATE_XML environment variable so they only run when explicitly requested. Previously these were ungated [Fact]s that rewrote all fixtures on every dotnet test, churning git and letting the comparison tests pass tautologically against output the parser had just produced, which masked regressions in shared code.

Changes
- Added a shared DocumentHelpers.SkipUnlessUpdatingFixtures guard that skips unless UPDATE_XML="true" is set, mirroring the existing judgment UpdateXmlFiles utility, and uses [CallerMemberName] so the skip message names the regenerator without hardcoded strings.
- Applied the guard to all 11 leg regenerators (RegenerateAllTestFiles for CoP/EM/EN/IA/OD/TN and RegenerateAllHtml for CoP/EM/EN/IA/TN).
- Added DocumentHelpers.WriteLegAknFixtureIfChanged so the .akn regenerators skip writing files whose only differences are non-deterministic metadata (FRBRdate dates, ukm:Parser version, uk:hash), keeping deliberate regenerations to meaningful diffs; the .html output embeds none of this, so it needs no equivalent.
- Gated the per-file "Regenerated …" log so it only prints for files that actually changed.
- Updated src/leg/README.md to document the UPDATE_XML regeneration workflow in place of the outdated "remove the Skip attribute" instructions.


test/leg/TestCoP.cs
```diff
@@ -85,10 +86,10 @@ public void RegenerateAllTestFiles() {
                 var result = Helper.Parse(docx, filename + ".docx", renderer: UK.Gov.Legislation.Test.LocalRendererHelper.GetOrNull());
                 var testFolder = System.IO.Path.Combine(projectRoot, "test", "leg", "cop");
                 var outputPath = System.IO.Path.Combine(testFolder, $"{filename}.akn");
-                System.IO.File.WriteAllText(outputPath, result.Serialize());
+                var written = DocumentHelpers.WriteLegAknFixtureIfChanged(outputPath, result.Serialize());
                 try { result.SaveImages(testFolder); }
```
dynamictulip: Question: Do you still want to save images if they have also not changed?

test/DocumentHelpers.cs
```diff
@@ -21,6 +22,46 @@ public static void AssertValidMainAkn(XmlDocument akn)
         Assert.Empty(UK.Gov.Legislation.Validator.Shared.ValidateAgainstMainAkn(akn));
     }
 
+    // Guards the fixture regenerators (RegenerateAllTestFiles / RegenerateAllHtml): they rewrite the
+    // golden fixtures on disk, so they only run when UPDATE_XML=true. Otherwise a normal test run
+    // rewrites every fixture and the comparison tests pass tautologically against their own output.
+    public static void SkipUnlessUpdatingFixtures([CallerMemberName] string caller = "")
```
dynamictulip: Thought (non-blocking): This method isn't really a DocumentHelper - it's not reading or manipulating the documents at all, it's controlling whether or not a "test" runs. Maybe a new file like `TestFileUpdateHelpers` would be good?

test/DocumentHelpers.cs
```diff
@@ -21,6 +22,46 @@ public static void AssertValidMainAkn(XmlDocument akn)
...
+    {
+        var enabled = bool.TryParse(Environment.GetEnvironmentVariable("UPDATE_XML"), out var update) && update;
+        Assert.SkipUnless(enabled,
+            $"Not a test. To run it: dotnet test test/test.csproj --filter \"FullyQualifiedName~{caller}\" -e UPDATE_XML=\"true\"");
+    }
+
+    // Strips non-deterministic metadata from a leg .akn (FRBRdate dates, ukm:Parser version, uk:hash)
+    // so a regeneration only diffs on real changes. (.html embeds none of this, so needs no equivalent.)
+    private const string LegNonDeterministicMetadataXslt = @"<?xml version='1.0'?>
```
dynamictulip: Question: I noticed that each leg akn test uses a different xslt to `RemoveSomeMetadata` before comparing (e.g. `IA` has a different one to `EN` and they're all different to this xslt) so I just want to double check - does this xslt cover all the reasons that you'd regenerate a file?

test/DocumentHelpers.cs
```diff
@@ -21,6 +22,46 @@ public static void AssertValidMainAkn(XmlDocument akn)
...
+  <xsl:template match='uk:hash/text()'/>
+  <xsl:template match='@*|node()'>
+    <xsl:copy>
+      <xsl:apply-templates select='@*|node()'/>
+    </xsl:copy>
+  </xsl:template>
+</xsl:stylesheet>";
+
+    // Writes the regenerated .akn only when it differs from disk after stripping non-deterministic
+    // metadata, keeping deliberate regenerations to meaningful diffs. Returns whether it wrote.
```
dynamictulip: Please can we use the triple slash xml doc comment format for this (and other overall class/method comments) because it will allow the intellisense to show this information when someone tries to use it

```suggestion
    /// <summary>
    /// Writes the regenerated .akn only when it differs from disk after stripping non-deterministic
    /// metadata, keeping deliberate regenerations to meaningful diffs. Returns whether it wrote.
    /// </summary>
```

test/leg/TestCoP.cs
QR-TNA: Good catch, no. Image-saving (and the log) now only run when the .akn actually changed, so unchanged files are left alone entirely.

test/DocumentHelpers.cs
QR-TNA: Agreed, moved both into a new TestFileUpdateHelpers class.

test/DocumentHelpers.cs
QR-TNA: Verified both ways: it strips a strict subset of what every per-type comparison XSLT strips (checked all six, they each strip these same three), so it can't miss a real change, only date/version/hash noise. And those three are the complete set of non-deterministic output: regenerating the whole .akn corpus today with UPDATE_XML=true writes 0 files, even though the committed fixtures are on an older date and parser version (a missed field would've rewritten everything).

dragon-dxw (approved):

PR #503, opened 2026-06-17
FCLC-2007: Backlog: Update consolidated logs script to work with tracker

dynamictulip:
This PR fixes the log consolidation script so it inserts Cloudwatch and Marklogic log csvs into the backlog parser tracker db (instead of expecting the old tracker csv file)


anthonyhashemi (approved):
Looks good! Approving assuming has been tested locally. _Could_ be worth a test once we stabilise this

dynamictulip:
We used it yesterday against staging - about to use now with prod!

PR #505, opened 2026-06-17
Upgrade to .NET 10

dragon-dxw:
.net 8 is EOL in November; we should upgrade to .net 10
https://national-archives.atlassian.net/browse/FCLC-1307


dragon-dxw:
Synk is apparently complaining about nothing:
image

dragon-dxw:
This is currently held back by a vulnerability in an SQLite package -- sit on it for a week or two before fixing and getting project-wide buy-in

dynamictulip (approved):

anthonyhashemi (approved):

PR #512, opened 2026-06-22
FCLC-1980: Save dummy dates from backlog spreadsheet

dragon-dxw:
https://national-archives.atlassian.net/browse/FCLC-1980

Ensure that we can enter dates of 1000-01-01 into the spreadsheet and they'll be recognised as dummy dates in the system.

See also #509 which I hadn't noticed at the time

### DIFF
--- a/test/XmlAssertions.cs
+++ b/test/XmlAssertions.cs
@@ -20,6 +20,14 @@ public static class XmlAssertions
         return nodeMatches[0]!;
     }
 
+    public static XmlNode HasAttribute(this XmlNode node, string attributeName, string expectedValue)
+    {
+        var attribute = node.Attributes?[attributeName];
+        Assert.NotNull(attribute);
+        Assert.Equal(expectedValue, attribute!.InnerText);
+        return node;
+    }
+
     public static void DoesNotHaveNodeWithName(this XmlDocument doc, string name)
     {
         var nodeMatches = doc.GetElementsByTagName(name);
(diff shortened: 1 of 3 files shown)


test/backlog/MetadataTests/TestStub.cs
```diff
@@ -205,4 +205,44 @@ public void Stub_WithNoWebArchivingLink_DoesNotAppearInXml()
...
+        var metadata = MetadataTransformer.MakeMetadata(line);
+
+        // Act
+        var stub = Stub.Make(metadata);
+        var xml = stub.Serialize();
+
+        // Assert
+        var doc = new XmlDocument();
+        doc.LoadXml(xml);
+        doc.HasSingleNodeWithName("FRBRWork").Which().HasChildWithName("FRBRdate").Which().HasAttribute("name", "dummy").And().HasAttribute("date", "1000-01-01");
```
anthonyhashemi, line 226: lovely

test/XmlAssertions.cs
```diff
@@ -135,7 +143,18 @@ public static XmlNode HasChildMatching(this XmlNode node, string expectedName, s
...
-    
+
+    public static XmlNode HasChildWithName(this XmlNode node, string name)
+{
+    var child = node.Cast<XmlNode>()
+        .FirstOrDefault(child =>
+            string.Equals(child.Name, name, StringComparison.InvariantCultureIgnoreCase));
+
+    if (child == null)
+        Assert.Fail($"Was not expecting to find a child of {node.Name} with name {name}");
```
anthonyhashemi: Shouldn't the message be the opposite?

"Expected to find a child of {node.Name} with name {name} but found none"

Do we need this message and logic at all though? If we want to be consitent with previous `Has..` Methods in this file we could just do:

`        Assert.NotNull(child);` like `HasSingleNodeWithName` and `HasAttribute`

anthonyhashemi (approved):

PR #538, opened 2026-07-02
lawmaker-sprint-13-fixes: Address all of Jim's comments, fixing various bugs

erciyass:


src/lawmaker/akn/Builder.cs
```diff
@@ -377,15 +377,17 @@ protected override XmlElement Block(XmlElement parent, ILine line, string name)
             if (styles.Count > 0)
                 block.SetAttribute("style", CSS.SerializeInline(styles));
             ContainingParagraphStyle = stripped.Style;
+            XmlElement tblock = doc.CreateElement("tblock", ns);
+            bool imgAdded = false;
             foreach (IInline inline in stripped.Contents)
                 // If line contains image, wrap with tblock first
-                if (inline is WImageRef wimageRef)
+                if (inline is WImageRef wimageRef && !imgAdded)
```
chatawayf, line 384: If two images are in the same `stripped.Contents`, then only the first image is added.
I tested briefly in Lawmaker and wasn't able to add multiple `p/img` tags under the same `tblock`.
I think we need to add a `tblock` per image?

src/lawmaker/akn/Builder.cs
chatawayf: This check is redundant with the added check below

src/lawmaker/akn/Builder.cs
```diff
@@ -499,19 +506,22 @@ private void ExtractDimensions(XmlElement img, IImageRef model)
         {
             foreach (string style in model.Style.Split(";"))
             {
-                img.SetAttribute(
+                if (style.Any(char.IsDigit))
```
chatawayf: I don't think this is adequate and may still throw in some circumstances. This is checking the key and the value not just the value.
I think `float.TryParse` is safer but this code is super fragile anyway.

chatawayf:
I've pushed some changes to address some of the comments I made. Looks good otherwise.
Merging...

PR #562, opened 2026-07-15
FCLC-2120: Refactor stub

dynamictulip:
The stub code and tests were getting difficult to reason about and manipulate - this is a refactor that removes duplication, fixes code warnings and makes things easier to read in both the stub and the stub tests.

There should be no semantic changes.

### DIFF
--- a/backlog/src/Stub.cs
+++ b/backlog/src/Stub.cs
@@ -1,3 +1,4 @@
+#nullable enable
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


backlog/src/Stub.cs
```diff
@@ -1,260 +1,260 @@
...
 
-        private void Meta(XmlElement judgment)
-        {
-            XmlElement meta = CreateAndAppend("meta", judgment);
-            Identification(meta);
-            Lifecycle(meta);
-            References(meta);
-            Proprietary(meta);
-        }
+        var author = CreateAndAppend("FRBRauthor", work);
```
anthonyhashemi, line 89: don't these 2 lines need to be in the else statement like before?

backlog/src/Stub.cs
```diff
@@ -1,260 +1,260 @@
...
-
-            foreach (var jurisdiction in Data.Jurisdictions)
-            {
-                XmlElement jurisdictionNode = CreateAndAppendUK("jurisdiction", proprietary);
-                proprietary.AppendChild(jurisdictionNode);
-                jurisdictionNode.AppendChild(Document.CreateTextNode(jurisdiction.ShortName));
-            }
-            
-            XmlElement year = CreateAndAppendUK("year", proprietary);
-            proprietary.AppendChild(year);
```
anthonyhashemi, line 176: where's this gone?

backlog/src/Stub.cs
```diff
@@ -1,260 +1,260 @@
...
-                proprietary.AppendChild(number);
-                number.AppendChild(Document.CreateTextNode(caseNo));
-
-            }
-            foreach (UK.Gov.NationalArchives.CaseLaw.Model.Party pty in Data.Parties)
-            {
-                XmlElement party = CreateAndAppendUK("party", proprietary);
-                proprietary.AppendChild(party);
-                party.SetAttribute("role", pty.Role.ShowAs());
-                party.AppendChild(Document.CreateTextNode(pty.Name));
```
anthonyhashemi, line 190: same here

backlog/src/Stub.cs
```diff
@@ -1,260 +1,260 @@
...
+        CreateAndAppendUk(proprietary, "sourceFormat", stubMetadata.SourceFormat);
+        CreateAndAppendUk(proprietary, "parser", Metadata.GetParserVersion());
+    }
 
-        public string Serialize() {
-            using MemoryStream stream = new ();
-            Serialize(stream);
-            return System.Text.Encoding.UTF8.GetString(stream.ToArray());
-        }
+    private XmlElement CreateAndAppendUk(XmlElement proprietary, string name, string value)
```
anthonyhashemi, line 231: I see - this has been changed so all those missing lines I commented about moved here

backlog/src/Stub.cs
```diff
@@ -1,260 +1,260 @@
...
+        proprietary.AppendChild(e);
 
+        e.AppendChild(document.CreateTextNode(value));
+        return e;
+    }
+
+    private void Body(XmlElement judgment)
+    {
+        CreateAndAppend("header", judgment);
+        var body = CreateAndAppend("judgmentBody", judgment);
```
anthonyhashemi, line 243: Lot's of instances in this PR where you have gone from explicit type variable declaration to just var. What's the reason for that?

anthonyhashemi: okay a quick read into var vs explicit types in dotnet since version 3 started to take me down a rabbithole which I had to stop myself from going down. If you think var is the way to go, I'm happy with that but if there are conventions like this we want the repo to adhere to, we should incorporate them into linting rules in pre-commit hooks

backlog/src/Stub.cs
dynamictulip: They weren't in the else statement before as C# ignores line breaks and indentation.
An else statement applies to the next block of code which is either a single statement (terminated with a `;`) or a group of statements enclosed in `{}`

backlog/src/Stub.cs
dynamictulip: It was already completely redundant because it's inside `CreateAndAppendUK`

backlog/src/Stub.cs
dynamictulip: I moved the text node definitions into `CreateAndAppendUk` because they were super duplicated and impairing readability

backlog/src/Stub.cs
dynamictulip: Already defined in the existing .editorconfig which the precommit will use - my local linter changed all these over.
Happy to have a chat about explicit types vs vars though if you'd like to know more about why I prefer them!

backlog/src/Stub.cs
anthonyhashemi: 😨

anthonyhashemi (approved):
Great refactor 👌

PR #563, opened 2026-07-15
FCLC-2120: Backlog: Fix dummy date stub outputs

dynamictulip:
This fixes some issues with dummy date handling in the Backlog parser.

## Changes

- Backlog docx
  - Ensure that the dummy date isn't passed to main parser code (because otherwise it will override the date pulled from the doc)
- Stub
  - Ensure that lifecycle isn't set for dummy dates
  - Use date name "dummy" in both FRBRWork and FRBRExpression (by setting the name when we create the stub metadata)

### DIFF
--- a/backlog/src/Stub.cs
+++ b/backlog/src/Stub.cs
@@ -77,14 +77,7 @@ internal class Stub
 
         var date = CreateAndAppend("FRBRdate", work);
         date.SetAttribute("date", stubMetadata.Date?.Date);
-        if (stubMetadata.Date?.Date == "1000-01-01")
-        {
-            date.SetAttribute("name", "dummy");
-        }
-        else
-        {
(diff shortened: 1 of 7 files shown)


backlog/src/Stub.cs
```diff
@@ -143,11 +136,16 @@ private void Manifestation(XmlElement identification)
 
     private void Lifecycle(XmlElement meta)
     {
+        if (stubMetadata.Date?.Date == Metadata.DummyDate)
+        {
+            return;
```
anthonyhashemi, line 141: It does not need a lifecycle element since there is no real source of this date in relation tothe lifecycle of document creation. ✅

anthonyhashemi (approved):
Makes sense!

PR #564, opened 2026-07-15
FCLC-2120: Backlog: Use cite for stub uk:year and add uk:number

dynamictulip:
This brings the stub behaviour more inline with the real parsed xml and stops the `uk:year` being 1000 when there is no decision date

## Changes

- Gets stub uk:year from ncn when available or falls back to decision year (if not dummy) when no valid ncn specified
- Adds uk:number to stub

### DIFF
--- a/backlog/src/MetadataTransformer.cs
+++ b/backlog/src/MetadataTransformer.cs
@@ -91,10 +91,21 @@ internal class MetadataTransformer(
     {
         var decisionDateAsString = line.DecisionDateTime.ToString("yyyy-MM-dd");
 
-        var wNamedDate = decisionDateAsString == UK.Gov.Legislation.Judgments.AkomaNtoso.Metadata.DummyDate
+        var isDummyDate = decisionDateAsString == UK.Gov.Legislation.Judgments.AkomaNtoso.Metadata.DummyDate;
+        var wNamedDate = isDummyDate
             ? new WNamedDate { Date = UK.Gov.Legislation.Judgments.AkomaNtoso.Metadata.DummyDate, Name = "dummy" }
             : new WNamedDate { Date = decisionDateAsString, Name = "decision" };
-        
+
+        var hasNcn = line.CleanedNcn is not null;
+        var shortUriComponent = hasNcn ? Citations.MakeUriComponent(line.CleanedNcn) : null;
+
+        var year = (hasNcn, isDummyDate) switch
+        {
+            (hasNcn: true, _) => Citations.YearFromUriComponent(shortUriComponent),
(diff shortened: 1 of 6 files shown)


backlog/src/MetadataTransformer.cs
```diff
@@ -91,10 +91,21 @@ internal static StubMetadata MakeMetadata(CsvLine line)
...
             ? new WNamedDate { Date = UK.Gov.Legislation.Judgments.AkomaNtoso.Metadata.DummyDate, Name = "dummy" }
             : new WNamedDate { Date = decisionDateAsString, Name = "decision" };
-        
+
+        var hasNcn = line.CleanedNcn is not null;
+        var shortUriComponent = hasNcn ? Citations.MakeUriComponent(line.CleanedNcn) : null;
+
+        var year = (hasNcn, isDummyDate) switch
+        {
+            (hasNcn: true, _) => Citations.YearFromUriComponent(shortUriComponent),
```
anthonyhashemi, line 104: we trust the ncn's year more often than the decision date's? or just an abitrary choice?

dynamictulip: I considered a few things when making this decision:

1. The real parser only ever takes the year from the NCN - if there is no NCN then there is no year
    - This makes sense because of the original assumption that all documents have NCNs
2. The current stub implementation only ever takes the year from the decision date
    - This makes sense because the backlog parser didn't used to deal with NCNs at all
3. The current stub implementation has to be updated in some manner otherwise we'll have a lot of pdf decisions with a year of `1000` from the dummy date

Our proprietary schema defines `year` as "The year of the judgment" which makes me think that both sources (NCN and decision date) are viable sources.

My options were to either:

1. Keep the existing stub behaviour of only looking at the decision date and then set year to null when there is no date.
    - This would mean that at a minimum, all batch 13 ,14 and 15 would have no `<uk:year>`
2. Copy the real parser behaviour of only looking at the NCN and then set year to null when there is no NCN.
    - This would mean that the majority of historic tribunals would have no `<uk:year>`
3. 🥇 Favour the NCN year over the decision date year
    - This is closest to real parser behaviour but still allows there to be a `<uk:year>` when there's no NCN
4. 🥈 Favour the decision date year over the NCN year
    - This is also acceptable in my opinion but as it's slightly further away from docx parser behaviour it lost out
    
In general, I don't trust the decision dates any less that I trust the NCNs in the external metadata

anthonyhashemi: thanks for clarifying that, and I agree with your choice given that!

anthonyhashemi (approved):

PR #578, opened 2026-07-21
Backlog - fix pdf ncn tracker

dynamictulip:
Some new e2e tracker tests found a bug - the ncns weren't being tracked in `ParserEvents` for PDFs. This PR fixes that and adds the first batch of e2e tracker tests.

# Changes

- Add e2e tracker db tests
- Record NCNs in the ParserEvents tracker table for PDFs

### DIFF
--- a/backlog/src/BacklogParserWorker.cs
+++ b/backlog/src/BacklogParserWorker.cs
@@ -14,7 +14,10 @@ using Backlog.Tracking;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
+using UK.Gov.Legislation.Judgments.AkomaNtoso;
+
 using Api = UK.Gov.NationalArchives.Judgments.Api;
+using Bundle = Backlog.Src.Bundle;
 
 namespace Backlog;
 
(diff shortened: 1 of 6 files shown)


backlog/src/BacklogParserWorker.cs
```diff
@@ -105,73 +108,71 @@ public async Task<int> RunAsync()
...
-
-            response = parser.Parse(request);
-
-            if (response.Xml.Contains("<header />"))
-            {
-                throw new NotSupportedException(
-                    "Couldn't parse header - try updating titles used to identify the end of header in OptimizedUKUTParser.titles");
-            }
+            throw new NotSupportedException(
+                "Couldn't parse header - try updating titles used to identify the end of header in OptimizedUKUTParser.titles");
```
anthonyhashemi, line 144: ah this is the code you were referring to around the header bug. Definitely will be worth adding this in in the normal parser flow when we take on that ticket!

backlog/src/BacklogParserWorker.cs
```diff
@@ -105,73 +108,71 @@ public async Task<int> RunAsync()
         return hasErrors ? 1 : 0;
     }
 
-    private Api.Response CreateResponse(CsvLine csvLine, string mimeType, byte[] sourceContent, bool isStub)
+    private Api.Response SendToParser(CsvLine csvLine, string mimeType, byte[] sourceContent)
```
anthonyhashemi, line 111: this change is an unrelated refactor right?

anthonyhashemi (commented):
I'm struggling to find the actual logic change that fixes the bug, could you point it out?

backlog/src/BacklogParserWorker.cs
dynamictulip: Kinda! I split this up into two so I could test the `MakeStubResponse` part

backlog/src/BacklogParserWorker.cs
```diff
@@ -105,73 +108,71 @@ public async Task<int> RunAsync()
...
+        return new Api.Response
+        {
+            Xml = stub.Serialize(),
+            Meta = new Api.Meta
+            {
+                DocumentType = "decision",
+                Court = stubMetadata.Court?.Code,
+                Date = stubMetadata.Date?.Date,
+                Name = stubMetadata.Name,
+                Cite = stubMetadata.Cite
```
anthonyhashemi, line 164: ah is it this line?

anthonyhashemi (approved):
Lovely stuff

PR #591, opened 2026-07-27
Backlog - move tests to separate assembly

dynamictulip:
Moves the backlog tests to a dedicated assembly

- ensure isolation by preventing accidental reference of backlog parser test code in non-backlog parser tests
- allow for easier separation of projects in the future
- allows stricter standards to be applied to just the backlog tests without affecting other teams

## Changes

- All backlog tests are in a dedicated project
- New linting rules have been applied to changed files
- Test cases are named by what they're testing rather than which test company is in the judgment (e.g. `Altaf Ebrahim t_a Ebrahim & Co v OISC` is now `doc`)
- Nullable context is enabled by default for all backlog tests so doesn't need to be specified per file


jacksonj04 (approved):
Aside from the tests going red (for no apparent reason?), :shipit:

dynamictulip:
> Aside from the tests going red (for no apparent reason?), :shipit:

(Hopefully) have fixed the tests - looked like a case sensitivity issue which was fine locally but less fine on CI

PR #603, opened 2026-08-06
feat(parser): Add First Tier Tribunal Immigration and Asylum

dynamictulip:
Adds UKFTT Immigration and Asylum Chamber to the UKUT parser.

## Changes

- Update courts.json to latest release (https://github.com/nationalarchives/ds-caselaw-utils/pull/900)
- Add ability for UKUT parser to identify that a document belongs to the UKFTT-IAC tribunal

### DIFF
--- a/src/parsers/ukut/Court.cs
+++ b/src/parsers/ukut/Court.cs
@@ -22,7 +22,7 @@ internal class CourtType : AbstractCourtType
 
         new GRCCombo()
     ];
-    
+
     protected override IEnumerable<Combo2> Combo2s { get; } =
     [
         new()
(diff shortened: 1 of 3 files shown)


src/parsers/ukut/Court.cs
```diff
@@ -46,6 +46,13 @@ internal class CourtType : AbstractCourtType
             Court = Courts.UpperTribunal_AdministrativeAppealsChamber
         },
 
+        new()
+        {
+            Re1 = Courts.FirstTierTribunalIdentifierRegex(),
+            Re2 = new Regex(@"^\(?Immigration and Asylum Chamber\)?$", RegexOptions.IgnoreCase),
```
dancs-dev, line 52: Just to double-check, this won't clash with the UKUT's Immigration and Asylum Chamber?

dancs-dev (approved):
LGTM. Could we re-run the pipeline now that the `courts.json` should be available?

src/parsers/ukut/Court.cs
dynamictulip: Should be good because of this bit ` Re1 = Courts.FirstTierTribunalIdentifierRegex(),`

PR #610, opened 2026-08-18
Lco 5093

palmer-bailey:


src/lawmaker/parsers/LegislationParser.cs
```diff
@@ -76,14 +80,21 @@ private LegislationParser(IEnumerable<IBlock> contents, DocumentStyle? style, Le
...
+            Logger.LogDebug($"Parsing header...");
+            ParseAndEnrichHeader();
+            Logger.LogDebug($"Parsing body...");
+            ParseBody();
+            Logger.LogDebug($"Parsing conclusions...");
+            ParseConclusions();
+        }
+        catch (Exception ex)
+        {
+            var blockText = (i >= 0 && i < Body.Count) ? Body[i]?.ToString() : null;
```
jurisdatum, line 95: `i` will not always be pointing to the problematic block, by the time this exception is caught. I can't really think of an easy way to improve. Making `i` a property (as the `TODO` at `BlockParser.cs:15` suggests) and caching `i` at the top of the ParseBody loop, might help. Not sure if that's worth doing now.

jurisdatum (approved):

src/lawmaker/parsers/LegislationParser.cs
palmer-bailey: Thanks for the feedback, this ticket was estimated with the assumption that the exception messaging would be a rough indication for the dev team/users. As we have other priorities at the moment I think we will stick with this approach for now and maybe circle back when we have more resources for word import development.

PR #612, opened 2026-08-19
test(parser): add party enricher unit tests

dynamictulip:
Add some unit tests to the Party Enricher to give a safety blanket before refactoring

### DIFF
--- a/test/parsers/common/enrich/TestPartyEnricher.cs
+++ b/test/parsers/common/enrich/TestPartyEnricher.cs
@@ -1,0 +1,611 @@
+using System.Linq;
+
+using DocumentFormat.OpenXml.Wordprocessing;
+
+using Shouldly;
+
+using UK.Gov.Legislation.Judgments;
+using UK.Gov.Legislation.Judgments.Parse;
+
+using Xunit;
+
+namespace test.parsers.common.enrich;
+
+public class TestPartyEnricher
+{
+    private static readonly PartyEnricher PartyEnricher = new();
+
+    private static readonly WLine LineTemplate = new(null, new Paragraph());
+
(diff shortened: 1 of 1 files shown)


test/parsers/common/enrich/TestPartyEnricher.cs
```diff
@@ -0,0 +1,611 @@
...
+    [InlineData("Interested Party", PartyRole.InterestedParty)]
+    [InlineData("Interested Parties", PartyRole.InterestedParty)]
+    [InlineData("Intervener", PartyRole.Intervener)]
+    [InlineData("Interveners", PartyRole.Intervener)]
+    [InlineData("Requested Person", PartyRole.RequestedPerson)]
+    [InlineData("Requesting State", PartyRole.RequestingState)]
+    [InlineData("Third Party", PartyRole.ThirdParty)]
+    [InlineData("Part 20 Defendant", PartyRole.Defendant)]
+    [InlineData("Claimant/Defendant", PartyRole.Claimant)]
+    [InlineData("Appellant/Respondent", PartyRole.Appellant)]
```
dancs-dev, line 60: Mainly for my own understanding of the legalese, aren't the appellant/respondent different parties?

dancs-dev (approved):
Looks good to me!

test/parsers/common/enrich/TestPartyEnricher.cs
dynamictulip: Not always...sometimes they can take on multiple roles which is why I believe the code currently has an ordered hierarchy of which to return.

See [here for examples](https://caselaw.nationalarchives.gov.uk/search?query=Appellant%2FRespondent)

PR #614, opened 2026-08-19
refactor(PartyEnricher): Fix multiple enumeration issues

dynamictulip:
Fixes multiple enumeration issues.

This ensures that any code that is run to create the contents of the line collections is not run each time we want to count the objects within that collection or to pluck on object out of it by forcing the collection to enumerate to an array. This improves performance, exposes better methods to analyse/manipulate the arrays and gives a step towards slimming down the parameters to methods.

### DIFF
--- a/src/parsers/common/enrich/PartyEnricher.cs
+++ b/src/parsers/common/enrich/PartyEnricher.cs
@@ -14,7 +14,7 @@ internal class PartyEnricher : Enricher
     internal override IEnumerable<IBlock> Enrich(IEnumerable<IBlock> blocks)
     {
         var before = blocks.ToArray();
-        var after = new List<IBlock>(before.Length);
+        var after = new List<IBlock>();
         var i = 0;
         while (i < before.Length)
         {
@@ -98,15 +98,12 @@ internal class PartyEnricher : Enricher
 
     private static List<IBlock> EnrichInTheMatterOf3(IBlock[] before, int i)
     {
-        var line1 = before[i];
-        var line2 = before[i + 1];
-        var line3 = before[i + 2];
-        var after = new List<IBlock>(3);
-        after.Add(line1);
-        var docTitle = MakeDocTitle(line2);
-        after.Add(docTitle);
-        after.Add(line3);
-        return after;
+        return
+        [
+            before[i],
+            MakeDocTitle(before[i + 1]),
+            before[i + 2]
+        ];
     }
 
     private static bool IsInTheMatterOf4(IBlock[] before, int i)
@@ -134,18 +131,13 @@ internal class PartyEnricher : Enricher
 
     private static List<IBlock> EnrichInTheMatterOf4(IBlock[] before, int i)
     {
-        var line1 = before[i];
-        var line2 = before[i + 1];
-        var line3 = before[i + 2];
-        var line4 = before[i + 3];
-        var after = new List<IBlock>(4);
-        after.Add(line1);
-        var docTitle1 = MakeDocTitle(line2);
-        var docTitle2 = MakeDocTitle(line3);
-        after.Add(docTitle1);
-        after.Add(docTitle2);
-        after.Add(line4);
-        return after;
+        return
+        [
+            before[i],
+            MakeDocTitle(before[i + 1]),
+            MakeDocTitle(before[i + 2]),
+            before[i + 3]
+        ];
     }
 
     /* three and four */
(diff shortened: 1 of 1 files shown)


src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1578,7 +1577,7 @@ protected override IEnumerable<IInline> Enrich(IEnumerable<IInline> line)
     private WTable EnrichTable(WTable table)
     {
         IEnumerable<WRow> rows = null;
-        if (table.TypedRows.Count() == 3)
+        if (table.TypedRows.Count == 3)
```
dancs-dev: Was the change from `.Count()` to `.Count` intentional? Just want to confirm there's no behaviour difference here.

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1578,44 +1549,32 @@ protected override IEnumerable<IInline> Enrich(IEnumerable<IInline> line)
...
```
dancs-dev, line 1552: Was the change from .Count() to .Count intentional? Just want to confirm there's no behaviour difference here.

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -2108,37 +2061,14 @@ private static WCell EnrichCellWithPartyRole(WCell cell, PartyRole role)
...
-        {
-            return null;
-        }
-
-        var block1 = lines.First();
-        if (block1 is not WLine line1)
-        {
-            return null;
-        }
+        var lines = cell.Contents.Where(l => !IsEmptyLine(l)).ToArray();
```
dancs-dev, line 2064: Out of curiosity, in the original, was there any reason it was iterated through twice? Was this needed:

```
        while (lines.Any() && IsEmptyLine(lines.Last()))
        {
            lines = lines.SkipLast(1);
        }
```

dancs-dev (approved):
Looks good to me! A very nice refactor

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: Yep, this is a difference between using the `Count` property on a list and the `Count()` function on a generic collection

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: There is a small difference in behaviour between the old and the new

Old steps:
1. Iterates through the collection and trims empty lines from the start (in reality it's actually creating a new collection on each loop)
2. Iterates through the collection and trims empty lines from the end (in reality it's actually creating a new collection on each loop)
3. Returns null if there's not two lines or if either of them are not wlines
4. Uses the values of those two lines to get the party role

New steps:
1. Removes all empty lines from anywhere in the collection in a single efficient pass
2. Checks that there's two lines and they are both wlines
  a. Uses the values of those two lines to get the party role
  b. Otherwise returns null

I can't see a scenario where it's bad to drop an empty line between two party roles and non of the e2e tests have failed

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1705,8 +1705,9 @@ private WRow EnrichRow(WRow row)
 
     private WRow EnrichRow2(WRow row)
     {
-        var first = (WCell)row.Cells.ElementAt(0);
-        var second = (WCell)row.Cells.ElementAt(1);
+        var rowCells = row.Cells.ToArray();
+        var first = (WCell)rowCells[0];
+        var second = (WCell)rowCells[1];
```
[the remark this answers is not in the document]
dynamictulip: Ignoring for now as this method will be folded into the caller in future

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -2108,37 +2103,14 @@ private static WCell EnrichCellWithPartyRole(WCell cell, PartyRole role)
...
-        var block2 = lines.Last();
-        if (block2 is not WLine line2)
-        {
-            return null;
-        }
-
-        var one = line1.NormalizedContent;
-        var two = line2.NormalizedContent;
-        return GetTwoLinePartyRole(one, two);
+        return null;
```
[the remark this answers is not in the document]
dynamictulip: Not important right now so ignoring

src/parsers/common/enrich/PartyEnricher.cs
dancs-dev: Thanks for clarifying!

PR #620, opened 2026-08-24
FCLC-431: Use Try pattern and extract some role mapping logic

dynamictulip:
The `PartyEnricher` had null returns and null handling throughout the class which obscured the intention of the methods and made it tricky to rely on the output of any given method. This PR transitions the `PartyEnricher` to the more standard "try" pattern where failure is explicitly indicated by the returned boolean being false.

Some methods have also been renamed and some of the role mapping logic has been extracted into static sets/dictionaries in order to make this transition to a try pattern slightly cleaner

### DIFF
--- a/src/parsers/common/enrich/PartyEnricher.cs
+++ b/src/parsers/common/enrich/PartyEnricher.cs
@@ -59,9 +59,9 @@ internal class PartyEnricher : Enricher
             }
 
             var rest = before[i..];
-            var found = EnrichMultiLinePartyBockOrNull(rest);
-            found ??= EnrichMultiLinePartyBockOrNull2(rest);
-            found ??= EnrichMultiLinePartyBockOrNull3(rest);
+            var found = EnrichMultiLinePartyBlockOrNull(rest);
+            found ??= EnrichMultiLinePartyBlockWithInlineRolesOrNull(rest);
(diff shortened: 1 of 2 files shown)


src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -316,365 +315,364 @@ private static List<IBlock> EnrichFiveLinePartyBlock(IBlock[] before, int i)
 
     /* multi-line */
 
-    private static List<IBlock> EnrichMultiLinePartyBockOrNull(IBlock[] rest, bool successive = false)
+    private static bool TryEnrichMultiLinePartyBlock(IBlock[] rest, bool successive, out List<IBlock> enriched)
```
anthonyhashemi, line 318: I was looking at StackOverflow and microsoft docs to read more about this pattern. 

Just flagging that this Try Pattern is different to the TryParse pattern I found  microsoft recommends here: https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/exceptions-and-performance#try-parse-pattern but that is specifically for dealing with exceptions more efficiently

Do you have any docs/links around the TryPattern you migrated to being more standard than handling nulls/defaults?

anthonyhashemi (approved):

anthonyhashemi (approved):
Talked through on slack. We previously had nulls when our typing didn't expect any, so this is a good step towards enabling nullabe references

PR #621, opened 2026-08-24
FCLC-431: Pass WLine rather than IBlock

dynamictulip:
Moves WLine checks higher up the caller tree so it can be checked in fewer places and so callers are more explicit about the expected state of the things that they're passing.

## Changes

- Removes duplicated `IsEmptyCell` method
- Inlines simple `EnrichBlock`
- Moves responsibility for ensuring things are `WLine` to callers of methods
- Inlines the different methods called by `TryGetPartyRole`


dancs-dev (approved):
Looks good to me

PR #625, opened 2026-08-26
refactor(PartyEnricher): replace common string checks with helper methods

dynamictulip:
Moves common string checks into helper methods which are responsible for their own trim checks. Introduces `IsNotBlank` alongside `IsBlank` for easier reading of negative cases

### DIFF
--- a/src/parsers/common/enrich/PartyEnricher.cs
+++ b/src/parsers/common/enrich/PartyEnricher.cs
@@ -764,13 +764,13 @@ internal class PartyEnricher : Enricher
         }
 
         if (lineContents.All(inline => inline is WText) &&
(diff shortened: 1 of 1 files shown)


src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -2025,16 +2025,41 @@ private WCell EnrichCell(WCell cell, PartyRole role)
         return new WCell(cell.Row, cell.Props, contents);
     }
 
+    private static bool IsNotBlank(WText wText)
```
anthonyhashemi, line 2028: my face when reading csharp _anymore_ 😛

anthonyhashemi (approved):
Luvly jubbly

PR #626, opened 2026-08-26
FCLC-431: Simplify `EnrichCell`

dynamictulip:
`EnrichCell` was an extremely difficult method to understand and contained unreachable code. This PR strips it down to its core, adding more unit tests to catch edgecases highlighted by the e2e tests and fixes an existing bug unveiled by these changes.

## Changes

- `EnrichCell` is renamed to `EnrichCellWithParty`
- The `EnrichCell` code concerned with `WLine` is moved to new `EnrichLineWithParty` and is cut back significantly
- `MakeOrSplitParty` becomes enveloped in new `EnrichWTextWithParties`
- Existing bug in party identification when there are multiple parties and one contains brackets in the name is fixed

### DIFF
--- a/src/parsers/common/enrich/PartyEnricher.cs
+++ b/src/parsers/common/enrich/PartyEnricher.cs
@@ -1582,23 +1582,15 @@ internal class PartyEnricher : Enricher
     {
         var lineContents = line.Contents.ToArray();
 
-        return lineContents switch
+        static bool IsPartOfPartyName(WText wText)
+            => IsNotBlank(wText)
+            && !IsInBrackets(wText.Text)
+            && !IsConnectorText(wText.Text);
+
+        return lineContents.OfType<WText>().Count(IsPartOfPartyName) switch
         {
-            [] => line,
-
-            [WText { Text: "SECRETARY OF STATE " }, WLineBreak, WText] // [2021] EWCA Civ 1876
-                => WLine.Make(line, [new WParty2(lineContents.Cast<ITextOrWhitespace>()) { Role = role }]),
-
-            _ when lineContents.OfType<WText>().Count(wText => IsNotBlank(wText)
-                    && !IsInBrackets(wText.Text)
-                    && !IsConnectorText(wText.Text)) == 1
-                => WLine.Make(line, lineContents.SelectMany(inline => EnrichWTextWithParties(inline, role)).ToArray()),
-
-            _ when lineContents.OfType<WText>()
-                               .Count(wText => IsNotBlank(wText) && !IsInBrackets(wText.Text)
-                                   && !IsConnectorText(wText.Text)) > 1
-                => WLine.Make(line, [new WParty2(lineContents.Cast<WText>()) { Role = role }]),
-
+            1 => WLine.Make(line, lineContents.SelectMany(inline => EnrichWTextWithParties(inline, role)).ToArray()),
+            > 1 => WLine.Make(line, [new WParty2(lineContents.Cast<ITextOrWhitespace>()) { Role = role }]),
             _ => line
         };
     }
--- a/src/parsers/common/enrich/PartyEnricher.cs
+++ b/src/parsers/common/enrich/PartyEnricher.cs
@@ -3,8 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
 
-using DocumentFormat.OpenXml.Wordprocessing;
-
 using UK.Gov.Legislation.Judgments.Utils;
 
 namespace UK.Gov.Legislation.Judgments.Parse;
@@ -855,23 +853,6 @@ internal class PartyEnricher : Enricher
         return false;
     }
 
-    private static IInline[] MakeOrSplitParty(string text, RunProperties props, PartyRole role)
-    {
-        if (text.StartsWith("(1)") && text.Contains("(2)"))
-        {
-            // ewhc/admin/2022/273
-            var i = text.IndexOf("(2)");
-            var text1 = text.Substring(0, i);
-            var text2 = text.Substring(i);
-            var party1 = new WParty(text1, props) { Role = role };
-            var party2 = new WParty(text2, props) { Role = role };
-            return [party1, party2];
-        }
-
-        var party = new WParty(text, props) { Role = role };
-        return [party];
-    }
-
     private static WLine MakeParty(WLine line, PartyRole? role)
     {
         var lineContents = line.Contents.ToArray();
@@ -1255,7 +1236,7 @@ internal class PartyEnricher : Enricher
             return new WRow(row.Table, row.TablePropertyExceptions, row.Properties,
             [
                 first,
-                EnrichCell(second, role),
+                EnrichCellWithParty(second, role),
                 EnrichCellWithPartyRole(third, role)
             ]);
         }
@@ -1293,7 +1274,7 @@ internal class PartyEnricher : Enricher
         {
             return new WRow(row.Table, row.TablePropertyExceptions, row.Properties,
             [
-                EnrichCell(first, role),
+                EnrichCellWithParty(first, role),
                 EnrichCellWithPartyRole(second, role)
             ]);
         }
@@ -1419,7 +1400,7 @@ internal class PartyEnricher : Enricher
             return new WRow(row.Table, row.TablePropertyExceptions, row.Properties,
             [
                 thisRowFirstCell,
-                EnrichCell(thisRowMiddleCell, role),
+                EnrichCellWithParty(thisRowMiddleCell, role),
                 thisRowLastCell
             ]);
         }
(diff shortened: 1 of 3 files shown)


test/judgments/test73.xml
```diff
@@ -10,7 +10,7 @@
           <FRBRauthor href="#ewhc-chancery-insolvencyandcompanies" />
           <FRBRcountry value="GB-UKM" />
           <FRBRnumber value="3056" />
-          <FRBRname value="THE JOINT ADMINISTRATORS OF LEHMAN BROTHERS HOLDINGS PLC v LEHMAN BROTHERS HOLDINGS INC." />
+          <FRBRname value="THE JOINT ADMINISTRATORS OF LEHMAN BROTHERS HOLDINGS PLC v LB GP NO.1 LIMITED (in Liquidation)" />
```
anthonyhashemi, line 13: e2e test changing, does it make sense when you look at the docs?

dynamictulip: This has changed because the rewrite of `EnrichCell` uncovered a bug with multiple party identification where those parties have brackets in their name which needed to be fixed for the rewrite to work.
This judgment has three parties but the first one (`LB GP NO.1 LIMITED (in Liquidation)`) wasn't being parsed correctly. Now it is being correctly identified as a party the case name has updated because it is the actual first respondent in this case

test/parsers/common/enrich/TestPartyEnricher.cs
```diff
@@ -618,4 +618,119 @@ public void Enrich_ThreeCellTableRow_WithRoleOnlyInFollowingRow_AssignsPartyFrom
...
+        var table = new WTable(null, null, null,
+        [
+            new WRow(TableOf(), null, null, [new WCell(RowOf(), null, [nameLine]), CellOf("Defendant")])
+        ]);
+
+        var result = PartyEnricher.Enrich([table]);
+
+        var resultTable = result.ShouldHaveSingleItem().ShouldBeOfType<WTable>();
+        var nameCellLine = resultTable.TypedRows.ShouldHaveSingleItem()
+                                      .TypedCells[0].Contents.ShouldHaveSingleItem()
```
anthonyhashemi, line 728: [Non blocking suggestion]: It feels like it _could_ be useful to abstract out this structure to be a test fixture that is in all these party tests, e.g.

Maybe as a function which takes in the party cell string of "Defendant", "Claimant", etc?

Or at least naming the variable something more explicit about its relationship, maybe `partyBlock` or `partyTable`

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1808,221 +1789,74 @@ private static readonly (Regex Pattern, PartyRole Role)[] NLinePartyRolePatterns
...
-                    return WLine.Make(line,
-                    [
-                        lineContents[0],
-                        lineContents[1],
-                        party
-                    ]);
-                }
+    private static IEnumerable<IInline> EnrichWTextWithParties(IInline inline, PartyRole role)
+    {
+        if (inline is WText text)
```
anthonyhashemi: Preferable to return early to reduce the nesting 
        if (inline is WText text) {
    return [inline];
   }

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1808,221 +1789,74 @@ private static readonly (Regex Pattern, PartyRole Role)[] NLinePartyRolePatterns
...
-                    }
+        return lineContents switch
+        {
+            [] => line,
 
-                    if (lineContents[2] is not WText wText3)
-                    {
-                        return line;
-                    }
+            [WText { Text: "SECRETARY OF STATE " }, WLineBreak, WText] // [2021] EWCA Civ 1876
```
anthonyhashemi: This is a very specific edge case, I assume there is no way to generalise this ?

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1808,221 +1789,74 @@ private static readonly (Regex Pattern, PartyRole Role)[] NLinePartyRolePatterns
...
+                => WLine.Make(line, [new WParty2(lineContents.Cast<ITextOrWhitespace>()) { Role = role }]),
 
-                    if (!Regex.IsMatch(wText1.Text, @"^\d\.$"))
-                    {
-                        return line;
-                    }
+            _ when lineContents.OfType<WText>().Count(wText => IsNotBlank(wText)
```
anthonyhashemi: Is it possible to break some of this out into some variables, as it is a little difficult to parse

anthonyhashemi: also, as an alternative, an early return if pattern might be easier to parse

e.g.

```
    if (lineContents.Length == 0)
    {
        return line;
    }

    if (IsSecretaryOfStateWrappedLine(lineContents))
    {
        var party = new WParty2(lineContents.Cast<ITextOrWhitespace>()) { Role = role };
        return WLine.Make(line, [party]);
    }
```

where

```
private static bool IsSecretaryOfStateWrappedLine(IInline[] lineContents)
{
    return lineContents is [WText { Text: "SECRETARY OF STATE " }, WLineBreak, WText];
}
```

etc.

anthonyhashemi: I did look into whether switch statements are particularly more performant than if statements and found this which suggests a rough guide of if there are 5 or less conditions, use whateve reads better
https://stackoverflow.com/questions/395618/is-there-any-significant-difference-between-using-if-else-and-switch-case-in-c

anthonyhashemi: ... big rabbit hole and lots of ifs (pun intended) buts and maybes.

It would be great to abstract out whatever possible though:

Something alone the lines of 

```
var lineContents = line.Contents.ToArray();

var isEmpty = lineContents.Length == 0;
var isSecretaryOfStateWrappedLine =
    lineContents is [WText { Text: "SECRETARY OF STATE " }, WLineBreak, WText];

var meaningfulTextCount = lineContents
    .OfType<WText>()
    .Count(wText => IsNotBlank(wText) && !IsInBrackets(wText.Text) && !IsConnectorText(wText.Text));

var hasSingleMeaningfulText = meaningfulTextCount == 1;
var hasMultipleMeaningfulText = meaningfulTextCount > 1;

return lineContents switch
{
    _ when isEmpty => line,
    _ when isSecretaryOfStateWrappedLine
        => WLine.Make(line, [new WParty2(lineContents.Cast<ITextOrWhitespace>()) { Role = role }]),
    _ when hasSingleMeaningfulText
        => WLine.Make(line, lineContents.SelectMany(inline => EnrichWTextWithParties(inline, role)).ToArray()),
    _ when hasMultipleMeaningfulText
        => WLine.Make(line, [new WParty2(lineContents.Cast<WText>()) { Role = role }]),
    _ => line
};
```

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1808,221 +1789,74 @@ private static readonly (Regex Pattern, PartyRole Role)[] NLinePartyRolePatterns
...
-                    var npContents = np.Contents.ToArray();
-                    if (npContents.Length != 1 || npContents[0] is not WText wText2)
-                    {
-                        return np;
-                    }
-
-                    var party2 = new WParty(wText2) { Role = role };
-                    return new WOldNumberedParagraph(np, [party2]);
-                }
+                    WOldNumberedParagraph wOldNumberedParagraph =>
```
anthonyhashemi: similar thoughts as the other switch statement

test/parsers/common/enrich/TestPartyEnricher.cs
```diff
@@ -618,4 +618,119 @@ public void Enrich_ThreeCellTableRow_WithRoleOnlyInFollowingRow_AssignsPartyFrom
                .Contents.ShouldHaveSingleItem().ShouldBeOfType<WRole>()
                .Role.ShouldBe(PartyRole.Claimant);
     }
+
+    [Fact]
```
anthonyhashemi: Amazing, reading these clear scenarios boosted my confidence in the implementation of the refactor (on top of the e2e tests)

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: We can totally do this but we we will still need to duplicate a `return [inline];` at the bottom of the method

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: Honestly I'm not sure, will investigate further

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: Not applicable here but one benefit of `switch` over `if` that I've been enjoying is that a var declared in a `switch` `case` pattern is only applicable for scope of that `case` whereas a var declared in an `if` condition is scoped beyond the reach of that `if`. Practically that means that vars in switches can reuse the same name whereas vars in chained if conditions have to be uniquely named

anthonyhashemi: that's pretty interesting, I wonder why there's that limitation in the if implementation!

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1577,221 +1558,68 @@ private static readonly (Regex Pattern, PartyRole Role)[] NLinePartyRolePatterns
...
+            && !IsConnectorText(wText.Text);
 
-                    if (!Regex.IsMatch(wText1.Text, @"^\d\.$"))
-                    {
-                        return line;
-                    }
+        return lineContents.OfType<WText>().Count(IsPartOfPartyName) switch
+        {
+            1 => WLine.Make(line, lineContents.SelectMany(inline => EnrichWTextWithParties(inline, role)).ToArray()),
+            > 1 => WLine.Make(line, [new WParty2(lineContents.Cast<ITextOrWhitespace>()) { Role = role }]),
```
anthonyhashemi, line 1603: particular reason why this isn't just `2` rather than `> 1`?

src/parsers/common/enrich/PartyEnricher.cs
anthonyhashemi: Might just be a personal preference thing so I don't want to impose, but what do you think about this?
Do you think it is easier to read? I do, and also feel like it would mean debugging would be simpler

```
private WBlock EnrichBlockWithParty(WBlock block, PartyRole role)
{
    switch (block)
    {
        case WOldNumberedParagraph p:
            return EnrichOldNumberedParagraphWithParty(p, role);

        case WLine line:
            return EnrichLineWithParty(line, role);

        default:
            return block;
    }
}

private WCell EnrichCellWithParty(WCell cell, PartyRole role)
{
    var contents = cell.Contents
        .Select(block => EnrichBlockWithParty(block, role))
        .ToArray();

    return new WCell(cell.Row, cell.Props, contents);
}

private static WOldNumberedParagraph EnrichOldNumberedParagraphWithParty(
    WOldNumberedParagraph paragraph,
    PartyRole role
)
{
    var parts = paragraph.Contents.ToArray();

    if (parts is [WText wText])
        return new WOldNumberedParagraph(paragraph, [new WParty(wText) { Role = role }]);

    return paragraph; // EWCA/Civ/2015/455
}
```

dynamictulip: I have been avoiding making more methods than necessary because there were so many similarly named partial duplicates of each other everywhere that I've been slowly wrangling into some sort of order 😅 
That said, I've a much better handle on it now and this is prettier so I've committed a slight variant.

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: There can legitimately be more than 2 parts I believe

test/parsers/common/enrich/TestPartyEnricher.cs
dynamictulip: I've merged together the ones that I can and simplified the test setup but the assertions are now all subtly distinct

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: Investigation indicated that it was surplus so have removed it

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: done

dancs-dev (approved):

PR #628, opened 2026-08-26
LCO-5042

leidos-lewys-ward:


chatawayf:
Hi @leidos-lewys-ward, happy to take a look at this but could you:
1. rebase to an updated main branch (currently based on a stale `main`) and
2. group the formatting commits together so that the diff can be viewed cleanly without the formatting changes. Maybe we should just run one big format on the lawmaker directory if it's causing a lot of headaches?

We could also look at adding the formatting commit(s) to a `.git-blame-ignore-revs` file so that we don't lose blame information.

leidos-lewys-ward:
> Hi @leidos-lewys-ward, happy to take a look at this but could you:
> 
> 1. rebase to an updated main branch (currently based on a stale `main`) and
> 2. group the formatting commits together so that the diff can be viewed cleanly without the formatting changes. Maybe we should just run one big format on the lawmaker directory if it's causing a lot of headaches?
> 
> We could also look at adding the formatting commit(s) to a `.git-blame-ignore-revs` file so that we don't lose blame information.

Thanks @chatawayf, these are good points so I've made the changes. I've rebased onto the current main, squashed the reformat commits into one, and added a .git-blame-ignore-revs file so blame still resolves through it. I diffed every file against the previous history afterwards and it's identical, then reran the full test suite and got the same result as before. I think I'd hold off from reformatting everything as that will expand the scope of the PR, but that might be worth doing in a future separate PR in one pass.

I think Neal has already assigned himself to the Jira ticket for review?

landryn-leidos (approved):

landryn-leidos:
I've reviewed the changes, confirmed all tests pass and confirmed the reported issues were fixed.  Everything looks good from here.  I don't think I can merge the changes myself as Sef or Francis are marked as reviewers.

src/docx/Numbering3.cs
chatawayf, line 94: The warning here can be fixed with a `#nullable enable`

leidos-lewys-ward: Resolved

landryn-leidos (approved):
Reverted changes look good and have resolved issue 1A

chatawayf:
We're awaiting @jurisdatum approval before merging into main, specifically for changes in modules outside `lawmaker`. The changes have been deployed to dev separately

src/docx/Numbering3.cs
```diff
@@ -7,437 +7,449 @@
...
+    {
+        Match match = paragraph.Descendants<Run>()
+            .Select(r => NamedListNumRegex().Match(r.InnerText))
+            .Where(m => m.Success)
+            .FirstOrDefault();
+        if (match is null)
+            return;
+
+        var name = match.Groups["name"].Value;
+        AbstractNum absNum = Numbering.GetAbstractNum(ctx.Main, name);
```
jurisdatum, line 371: Could you check the case where the newly matched unquoted name has no corresponding `<w:name>` in `numbering.xml`? `GetAbstractNum` can return null here and the next line dereferences it. Built-ins like `LegalDefault` can take that path when they have no corresponding named abstract definition, as in `test49.docx`.

Since a paragraph can hold several `LISTNUM` fields, I think this wants to select the first resolvable match rather than return on the first unresolved one. `test49.docx` has `LISTNUM LegalDefault \l 2` immediately followed by `LISTNUM "SEQ1" \l 2 \s 1` in the same paragraph, so which one wins here has changed. That paragraph’s style carries no numbering, so it doesn’t actually reach this code today.

src/parsers/common/fields/ListNum.cs
```diff
@@ -9,223 +9,259 @@
...
-        int fieldLvl = -1;
+        var fieldLvl = -1;
         match = Regex.Match(fieldCode, @"\\l (\d)");
-        if (match.Success) {
+        if (match.Success)
+        {
             fieldLvl = int.Parse(match.Groups[1].Value) - 1; // fieldLvl is 0 based
         }
-        int fieldStart = -1;
+        else if (Regex.IsMatch(fieldCode, @"\\l \(\d+\)"))
```
jurisdatum, line 168: Could you check an unresolved name combined with `\l (2) \s 1`? This new branch sends it to `FormatNumber(name, …)`, which dereferences a missing abstract-number definition and throws. Previously the catch-all returned a plain number; a guard falling back to that would keep its stated purpose of not throwing.

jurisdatum (commented):
Do you have a source-derived fixture with the parenthesized `\l (N)` form? `field-codes.docx` covers the unquoted/double-spaced `seq1` case, but I couldn’t find a parenthesized operand in any committed fixture. That means the `\l (N)` handling, including the behavior restored by `064d9ff1`, isn’t pinned by anything in the repository. It might be worth preserving the motivating field shape as a regression test.

src/parsers/common/fields/ListNum.cs
leidos-lewys-ward: Updated Numbering3.HandleNamedListNum so that it now iterates all LISTNUM matches in a paragraph and picks the first one that actually resolves via GetAbstractNum, rather than blindly dereferencing the first match.

leidos-lewys-ward:
Hi @jurisdatum,

I've addressed your points in commit fa37c51 as follows:

> Do you have a source-derived fixture with the parenthesized `\l (N)` form? `field-codes.docx` covers the unquoted/double-spaced `seq1` case, but I couldn’t find a parenthesized operand in any committed fixture. That means the `\l (N)` handling, including the behavior restored by `064d9ff1`, isn’t pinned by anything in the repository. It might be worth preserving the motivating field shape as a regression test.

I've added a regression paragraph to test/lawmaker/wsi/field-codes.docx/.xml using the real-world field code (LISTNUM  seq1 \l (2) \s 1) I found in the actual Springwell Solar DCO test document (see LCO-5042 jira ticket), appended as a new sibling paragraph tag at the end of article 8 so it doesn't disturb any existing numbering/structure.

> Could you check the case where the newly matched unquoted name has no corresponding <w:name> in numbering.xml? GetAbstractNum can return null here and the next line dereferences it. Built-ins like LegalDefault can take that path when they have no corresponding named abstract definition, as in test49.docx.

> Since a paragraph can hold several LISTNUM fields, I think this wants to select the first resolvable match rather than return on the first unresolved one. test49.docx has LISTNUM LegalDefault \l 2 immediately followed by LISTNUM "SEQ1" \l 2 \s 1 in the same paragraph, so which one wins here has changed. That paragraph’s style carries no numbering, so it doesn’t actually reach this code today.

Updated Numbering3.HandleNamedListNum so that it now iterates all LISTNUM matches in a paragraph and picks the first one that actually resolves via GetAbstractNum, rather than blindly dereferencing the first match.

> Could you check an unresolved name combined with \l (2) \s 1? This new branch sends it to FormatNumber(name, …), which dereferences a missing abstract-number definition and throws. Previously the catch-all returned a plain number; a guard falling back to that would keep its stated purpose of not throwing.

I've updated ListNum.CatchAllListNums to guard the FormatNumber(name, …) call behind a resolution check, falling back to "(1)" (matching the existing no-name behavior) instead of letting the function throw an exception.

jurisdatum (approved):
Great work. Well done

PR #630, opened 2026-08-28
refactor(PartyEnricher): make party-role string comparisons case-insensitive

dynamictulip:
Part of the complexities in the party role identification is due to multiple entries for different ways to case the party roles, e.g. "APPELLANT" vs "Appellant" vs "appellant". This PR switches to using case insensitive lookups for party roles and removes the duplicated entries

### DIFF
--- a/src/parsers/common/enrich/PartyEnricher.cs
+++ b/src/parsers/common/enrich/PartyEnricher.cs
@@ -5,6 +5,8 @@ using System.Text.RegularExpressions;
 
 using DocumentFormat.OpenXml.Wordprocessing;
 
+using UK.Gov.Legislation.Judgments.Utils;
+
 namespace UK.Gov.Legislation.Judgments.Parse;
 
 // there are "third" paries in EWCA/Civ/2015/631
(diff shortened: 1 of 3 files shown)


src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -130,28 +132,9 @@ private static List<IBlock> EnrichInTheMatterOf4(IBlock[] before, int i)
...
-
-        if (content == "R E G I N A")
-        {
-            return true;
-        }
-
-        return false;
+        var content = Regex.Replace(line.NormalizedContent, @"\s+", "");
+        return content.Equals("REX", StringComparison.OrdinalIgnoreCase)
+            || content.Equals("REGINA", StringComparison.OrdinalIgnoreCase);
```
dancs-dev, line 137: I think Rex and Regina could be names - could making this case-insensitive cause issues?

dancs-dev (approved):

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: It'd be highly unlikely as the individual would also need to not have a surname as these are checking the entire content of the run. On a side note, there may be a way to remove them altogether in the near future...

PR #633, opened 2026-09-01
FCLC-431: Simplify party role identification

dynamictulip:
This PR merges together the different ways to determine party roles into the `TryGetPartyRole` method and adds some more unit tests to this area.

Rather than using multiple competing sets of string comparisons to find the primary role for a party with multiple roles, the code now finds the role candidates for each part of a multi-role string and does a simple switch to determine the overall winning role.

### DIFF
--- a/src/api/Parser.cs
+++ b/src/api/Parser.cs
@@ -95,6 +95,20 @@ public class Parser(ILogger<Parser> logger, AkN.IValidator validator) : IParser
             var preParsed = new PreParser().Parse(doc);
             var attach2 = AkN.Parser.ConvertAttachments(attachments);
             IJudgment judgment = f(doc, preParsed, meta, attach2);
+
+            //** HACK - suppress misidentified parties by calling `CaseName.Extract` **
+            // For KBD and QBD cases, the king/queen and "on the application of" pieces of text are misidentified as
+            // their own parties.
+            // On a clean parse of the document, we create a case name which uses these misidentified parties to decide
+            // whether or not to add "(R on the application of)" to the case name and then "suppresses" these
+            // misidentified parties so they are not output in the xml.
+            // On a reparse of the document we are supplied with a case name via the outside metadata so we don't
+            // attempt to generate one. This means that the misidentified king/queen and "on the application of" parties
+            // are never suppressed on reparse and so incorrectly end up in the final xml.
+            // This hack fixes this issue by ensuring that the case name generation (and thus the misidentified party
+            // suppression) is always called.
+            _ = CaseName.Extract(judgment);
+
             return new AkN.Bundle(doc, judgment);
         };
     }
--- a/src/api/Parser.cs
+++ b/src/api/Parser.cs
@@ -1,35 +1,31 @@
-
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Xml;
 using System.Xml.Schema;
-
-using DocumentFormat.OpenXml.Packaging;
 
 using Microsoft.Extensions.Logging;
 
 using UK.Gov.Legislation.Judgments;
-using UK.Gov.NationalArchives.CaseLaw.Parse;
-using AkN = UK.Gov.Legislation.Judgments.AkomaNtoso;
-using PS = UK.Gov.NationalArchives.CaseLaw.PressSummaries;
-
-using AttachmentPair = System.Tuple<byte[], UK.Gov.Legislation.Judgments.AttachmentType>;
-using ParseFunction = System.Func<byte[], UK.Gov.Legislation.Judgments.IOutsideMetadata, System.Collections.Generic.IEnumerable<System.Tuple<byte[], UK.Gov.Legislation.Judgments.AttachmentType>>, UK.Gov.Legislation.Judgments.AkomaNtoso.ILazyBundle>;
-
-using AttachmentPair1 = System.Tuple<DocumentFormat.OpenXml.Packaging.WordprocessingDocument, UK.Gov.Legislation.Judgments.AttachmentType>;
-using OptimizedParseFunction = System.Func<DocumentFormat.OpenXml.Packaging.WordprocessingDocument, UK.Gov.NationalArchives.CaseLaw.Parse.WordDocument, UK.Gov.Legislation.Judgments.IOutsideMetadata, System.Collections.Generic.IEnumerable<System.Tuple<DocumentFormat.OpenXml.Packaging.WordprocessingDocument, UK.Gov.Legislation.Judgments.AttachmentType>>, UK.Gov.Legislation.Judgments.Parse.Judgment>;
 using UK.Gov.Legislation.Judgments.Parse;
+using UK.Gov.NationalArchives.CaseLaw.Parse;
+
+using AkN = UK.Gov.Legislation.Judgments.AkomaNtoso;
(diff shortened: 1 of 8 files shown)


src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1408,84 +1311,131 @@ private WRow EnrichRowWithPartyRoleFromNextRow(WRow row, WRow next)
...
+    {
+        var lineContents = cell.Contents
+                               .OfType<WLine>()
+                               .Where(LineHasContent)
+                               .Select(l => l.NormalizedContent)
+                               .ToArray();
+        return TryGetPartyRole(out role, lineContents);
+    }
+
+    public static bool TryGetPartyRole(out PartyRole role, params string[] inputRoleStrings)
```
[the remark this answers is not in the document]
dynamictulip: False alarm, this has no usages outside this repo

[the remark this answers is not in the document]
anthonyhashemi: is the comment about the TryParse/TryGet* convention valid though @dynamictulip ?

test/parsers/common/enrich/TestPartyEnricher_TryGetSinglePartyRole.cs
```diff
@@ -0,0 +1,210 @@
...
+    [InlineData("1st Claimant", PartyRole.Claimant)]
+    [InlineData("Jane Doe", null)]
+    public void TryGetPartyRole_Cell_WithOneNonEmptyLine_ParsesRoleFromCellText(string text, PartyRole? expected)
+    {
+        var cell = CellOf(text);
+
+        var found = PartyEnricher.TryGetPartyRole(cell, out var actual);
+
+        found.ShouldBe(expected is not null);
+        if (expected is not null)
```
anthonyhashemi: I know this already existed but since we're here I suggest we avoid conditionals in tests so that each test is testing one scenario (albeit with various permutations) for both readability, debugabillity and documentation.

So here I would move `    [InlineData("Jane Doe", null)]` into its own test as it is a distinct scenario from the others

src/parsers/common/enrich/PartyEnricher.cs
anthonyhashemi: I assume this pattern of returning a bool whilst defining an argument you will output by mutating it without explicit return is standard?

anthonyhashemi: ah yes its that Try** pattern again

src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -1408,84 +1311,131 @@ private WRow EnrichRowWithPartyRoleFromNextRow(WRow row, WRow next)
...
-                return true;
+        var cleanedParts = parts.Select(p => p.CleanWhitespace()
+                                              .Trim('(', ')')
+                                              .Trim());
 
-            case [var one, var two] when TwoLinePartyRoles.TryGetValue((one, two), out role)
-                || TryGetPartyRoleForCombinedLabels(one, two, out role):
-                return true;
+        var cleanedPartsWithoutPrefixes = cleanedParts
+                                          .Select(p => p.Equals("Third Party", StringComparison.OrdinalIgnoreCase)
```
anthonyhashemi: Could extract the condition to `ShouldSkipPrefixStripping` or `SemanticRoleLabel` or similar to make the intention clearer

anthonyhashemi (approved):
Great refactor. A couple of small non blocking suggestions/queries. Approved assuming the integration test issues are fixed

src/parsers/common/enrich/PartyEnricher.cs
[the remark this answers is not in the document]
dynamictulip: It is a bit weird to have the out parameter come first rather than last but necessary if I want to use the params keyword. I could circumvent this by creating another overload of the method - will have a fiddle

src/parsers/common/enrich/PartyEnricher.cs
dynamictulip: Upon reflection, I reckon this condition should belong inside the prefix stripping method rather than here

test/parsers/common/enrich/TestPartyEnricher_TryGetSinglePartyRole.cs
dynamictulip: Good spot!
This test content is actually already covered by `TryGetSinglePartyRole_ParsesRoleFromFreeText` and `TryGetSinglePartyRole_ReturnsFalseWhenNotRole` so I've removed it.

dynamictulip: Hold on, that's a lie - this is looking at it from a cell viewpoint.... I'll put it back in

dynamictulip: I've split it out and added a few extra cell tests

test/TRE/TestNoChange.cs
```diff
@@ -16,11 +14,10 @@ namespace test.TRE;
 
 public class TestNoChange
 {
-    private static readonly int[] FailingTestsToSkip = [6, 8, 31, 32];
```
anthonyhashemi: No more skipped tests!

However, could we move these cases to be a distinct test so it is obvious they only work due to the hack

src/api/Parser.cs
```diff
@@ -95,6 +95,16 @@ private static ParseFunction Wrap(OptimizedParseFunction f)
             var preParsed = new PreParser().Parse(doc);
             var attach2 = AkN.Parser.ConvertAttachments(attachments);
             IJudgment judgment = f(doc, preParsed, meta, attach2);
+
+            // ** HACK - Extract case name after judgment has been built in order to suppress parties on reparse **
+            //   For KBD and QBD cases, the king/queen and "on the application of" pieces of text are misidentified as
+            //   parties. This is then used by the CaseName extractor to build up the case name of a judgment. The
+            //   CaseName extractor then "suppresses" these parties so they are not output in the xml.
```
anthonyhashemi: little confused. I assume line 102 is talking about `CaseName.Extract` on line 106 whereas 101 is talking about `CaseNumber.Extract` occurring somewhere earlier in the reparse journey

dynamictulip: This is a super confusing scenario but I don't want the comment adding to the confusion! Does this rephrasing work?

** HACK - suppress misidentified parties by calling `CaseName.Extract`**
For KBD and QBD cases, the king/queen and "on the application of" pieces of text are misidentified as   their own parties.
On a clean parse of the document, we create a case name which uses these misidentified parties to decide whether or not to add "(R on the application of)" to the case name and then "suppresses" these misidentified parties so they are not output in the xml.
On a reparse of the document we are supplied with a case name via the outside metadata so we don't attempt to generate one. This means that the misidentified king/queen and "on the application of" parties are never suppressed on reparse and so incorrectly end up in the final xml.
This hack fixes this issue by ensuring that the case name generation (and thus the misidentified party suppression) is always called.

anthonyhashemi: Yes, that does! Thank you

test/TRE/TestNoChange.cs
dynamictulip: I've moved them to the parser tests rather than the TRE ones so they're closer to where the hack actually is!

anthonyhashemi (approved):

PR #639, opened 2026-09-08
FCLC-431: Reduce complexity of some methods

dynamictulip:
Refactored a number of methods with high cognitive complexities to make them more maintainable

## Changes

- `EnrichPartyNamesWithTwoRoles` - condensed into a switch with pattern matching
- `EnrichPartyTypesWithTwoRoles` - reordered to remove a `continue`
- `IsPartyName` - most of the code in this method was unreachable


src/parsers/common/enrich/PartyEnricher.cs
```diff
@@ -584,123 +584,8 @@ private static bool IsPartyName(WLine line)
...
-            if (IsNotBlank(wText2))
-            {
-                return false;
-            }
-
-            return true;
-        }
-
-        return false;
+        return lineContents.All(inline => inline is ITextOrWhitespace) &&
```
anthonyhashemi, line 587: wow, that was a simplification and a half!

anthonyhashemi (approved):
Lovely stuff
Private methods so no need to test directly 👌 
Would be cool to add a code coverage tool just to have an understanding if these paths are covered without having to go digging through codebase on each pr

PR #644, opened 2026-09-09
FCLC-2255: Support UKFTT-IAC anonymity directions

dynamictulip:
The new UKFTT-IAC court contains anonymity directions in the party block which were being misidentified as people. This PR ensures that they are excluded from party identification when the parties are supplied in plain WLine blocks. It does not handle anonymity directions in a table because that hasn't yet been seen and it is unclear what it would look like

## Changes

- Anonymity directions are no longer misidentified as parties
- A new e2e test comprised of a stripped back fake UKFTT-IAC decision have been added


dancs-dev (approved):
Looks good to me!

